PR #11, opened 2014-12-31
Add symmetry constraints for SCS

### LINKED ISSUE #8: Add symmetry constraints for SDP
The MPB interface implicitly assumes that SDP constraints imply symmetry, but SCS needs these as explicit equalities, see discussion here: https://github.com/JuliaOpt/Mosek.jl/issues/29#issuecomment-66731064. These equalities should be introduced in SCS.jl and not by Convex.jl.

karanveerm:
All Convex.jl tests also continue to pass. Fixes #8. Can be made more efficient.

### DIFF
--- a/src/SCSSolverInterface.jl
+++ b/src/SCSSolverInterface.jl
@@ -331,6 +331,46 @@ loadconicproblem!(model::SCSMathProgModel, c, A, b, constr_cones, var_cones) =
     loadconicproblem!(model, c, sparse(A), b, constr_cones, var_cones)
 
 
+# TODO: At least for Convex.jl, we are guaranteed what A and b look like
+# This should allow us to make it far more efficient
+
+# Given an n x n matrix vectorized in the form Ax - b where x is an unknown variable
+# Let y = b - Ax and Y = reshape(y, n, n)
+# This function then returns a matrix G and vector h
+# s.t Gx + s == h, s in the zero cone enforces Y[i, j] == Y[j, i]
+function addsymmetricityconstraints(A, b)
+    size_b = length(b)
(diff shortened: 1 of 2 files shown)


karanveerm:
The tests are passing on Windows, but not on OSX or Linux.
```
/home/travis/julia/bin/julia: symbol lookup error: /home/travis/.julia/v0.3/SCS/deps/usr/lib/libscsdir.so: undefined symbol: dsyevr_64_
```
dysevr seems to be an LAPACK function (http://phase.hpcc.jp/mirrors/netlib/lapack/double/dsyevr.f)

@mlubin Any ideas? Turns out the only previous SDP test is also commented out https://github.com/JuliaOpt/SCS.jl/blob/master/test/direct.jl#L117 and only SDPs use BLAS/LAPACK.

mlubin:
Oh, that's an unfortunate hole in the test coverage! I'll take a look.

mlubin:
I reset travis and it seems okay now.

joehuchette:
In keeping with the [MPB conic interface](http://mathprogbasejl.readthedocs.org/en/latest/conic.html), shouldn't this be handled at at the Convex.jl level instead?

mlubin:
I don't think SCS implements the new lower triangle SDP format yet... this should change.

bodono:
SCS has supported the lower triangle SDP format for a while, since this commit 9188912c7eb0bf80b796fbef6ce8235ae2437d25. I don't think the bundled up SCS packages contain that change yet though.

mlubin:
@bodono, ah yes, I mean the wrapper. Is the lower triangle SDP format on the master branch of SCS now?

bodono:
Yep, the master branch of the core SCS project uses the lower triangular SDP format.

PR #12, opened 2015-01-01
explicitly load lapack with RTLD_GLOBAL

mlubin:
We hadn't been testing the SCS functionality which calls lapack, oops.
@tkelman @staticfloat, is this a reasonable fix? Not sure how else to enable SCS to find julia's lapack symbols. It should be safe to load these globally now that the 64-bit integer API won't cause name clashes.
CC @karanveerm

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -2,7 +2,13 @@ using BinDeps
 
 @BinDeps.setup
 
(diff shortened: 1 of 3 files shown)


mlubin:
Travis passes, so that's a strict improvement.

PR #19, opened 2015-03-01
Update max_iters + fix deprec warning

karanveerm:
cc @madeleineudell @davidlizeng @jennyhong 

Throughout [EE364A](http://web.stanford.edu/class/ee364a/), we have been running into issues where the problem solution is inaccurate because max_iters has been too small (and also outside of EE364A: https://github.com/JuliaOpt/Convex.jl/issues/64 or [time series](http://nbviewer.ipython.org/github/JuliaOpt/Convex.jl/blob/master/examples/time_series/time_series.ipynb)).

Prof. Boyd really wants us to make `max_iters` 20k by default and change `eps` to `1e-4`. 

Also get rid of deprec warning.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.3
 MathProgBase
 BinDeps
 @osx Homebrew
+Compat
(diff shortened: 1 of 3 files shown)


bodono:
It would be useful to collect a list of problems that SCS takes a long time to converge on, so we have a suite of test problems to test performance improvements on. Do you know which problems from the class were challenging for SCS?

karanveerm:
@bodono sounds good. @jennyhong can you email him the problems where SCS.jl took more than 2.5k iterations?

PR #21, opened 2015-03-22
shouldn't need to do Pkg.checkout("MathProgBase") any more

tkelman:
pretty sure there's been a tag since this was added


mlubin:
AppVeyor is hanging on the nightlies. Any ideas why?

tkelman:
Don't know. If I had to guess, the deprecations warnings might be slowing things down inordinately in some Windows-specific and/or 0.4-specific way?

mlubin:
Merging anyway since it's a strict improvement. Thanks for the fix!

PR #28, opened 2015-06-18
update test and readme

bodono:
Hopefully tests should pass now. Minor change to readme too.


bodono:
The OSX tests are failing because it's still download 1.0.7.

One of the linux tests is failing, but one is passing. The one that's failing has
```
ERROR: LoadError: UndefVarError: nidx not defined
 in orderconesforscs at /home/travis/.julia/v0.4/SCS/src/SCSSolverInterface.jl:253
```

but that line is just

```
252             nidx = length(idxs)
253             A = [A; -sparse(1:nidx, idxs, ones(nidx), nidx, num_vars)]
```

How could that be failing like that?

mlubin:
Pretty sure this is a codegen error. Will file a julia issue.

mlubin:
Actually it looks like the wrong line number was reported. The issue is fixed with https://github.com/JuliaOpt/SCS.jl/commit/71df8d291a986f00dad1b645c56e469cb0432f2b

bodono:
Did the stack trace report the wrong number or was I looking at an old version?

@mlubin I don't understand how your change fixed it?

In any case, should we merge this in now then?

mlubin:
@bodono, ``nidx`` was only defined in the first case of the ``if`` but not inside the ``else``

mlubin:
I restarted travis, and this is passing on 0.3 and 0.4 on linux.

PR #29, opened 2015-06-24
Use v1.1.5 and update windows binary checksums accordingly

tkelman:
note - not tested locally yet, just built and uploaded so far

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -15,7 +15,7 @@ scs = library_dependency("scs", aliases=aliases)
     provides(Homebrew.HB, "scs", scs, os = :Darwin)
 end
 
-version = "1.1.1" # how do we give a version range?
+version = "1.1.5" # how do we give a version range?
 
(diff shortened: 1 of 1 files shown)


bodono:
Ship it!

mlubin:
Will people who already have SCS installed get updated to the new binaries with ``Pkg.update()``? It looks like they won't?

tkelman:
No, I don't think that will happen automatically. Even `Pkg.build` might not force a re-download / re-build I think, unless maybe there's a validation routine that we can use to distinguish between the two versions of the library?

mlubin:
@bodono, does SCS export a function to ask for the compiled version?

tkelman:
I do see an `scs_version` exported from the dll

bodono:
Right, there's a c function `scs_version` that returns a string, e.g. "1.1.5" right now, but it's not being exposed though SCS.jl right now. Should I add it?

bodono:
PR here: https://github.com/JuliaOpt/SCS.jl/pull/33

Issue #31, opened 2015-06-26
Julia crash on tiny change


bodono:
Changing three lines https://github.com/JuliaOpt/SCS.jl/blob/master/test/options.jl#L45 from
```
45 primal_sol = m.primal_sol
46 dual_sol = m.dual_sol
47 slack = m.slack
```
to
```
45 primal_sol = deepcopy(m.primal_sol)
46 dual_sol = deepcopy(m.dual_sol)
47 slack = deepcopy(m.slack)
```

results in 
```
julia(32256,0x7fff7b9f6300) malloc: *** error for object 0x7faee77bb0a0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug

signal (6): Abort trap: 6
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
```

my guess is that this is in julia itself, but I can't tell, could also be in MathProgBase or SCS.jl. @mlubin do you have any insight on where this could be coming from?

bodono:
I should mention I'm on OSX 10.10.3, 64 bit, using the standard homebrew SCS 1.1.5.

mlubin:
@bodono, does SCS ever call ``free()`` on memory passed to it?

mlubin:
Where did the memory [here](https://github.com/bodono/SCS.jl/blob/3d86eef52a73ae163d1f046850ee7fa462d4d553/src/high_level_wrapper.jl#L141) come from? Why is it converted to a pointer and then back to an array?

bodono:
It never calls free on memory passed in.

I don't know where that memory came from. Looks like it used to be
```
    x = copy(pointer_to_array(ptr_x, n))
    y = copy(pointer_to_array(ptr_y, m))
    s = copy(pointer_to_array(ptr_s, m))
```

should we change it back to that?

mlubin:
Does SCS allocate that memory? Is there a way to free it from inside SCS?

mlubin:
It's definitely safer to copy the memory into a julia array instead of asking julia to take over the garbage collection for memory which it didn't allocate.

bodono:
Yes, in this case it is allocated inside of SCS, because SCS checks if the memory is there before and if not then allocates it.

I can change it back then in another PR. Why does julia allow that operation if it's unsafe then?

bodono:
But if that memory is not cleared out by julia then it will result in a memory leak.

mlubin:
[...]

PR #32, opened 2015-06-26
add power cone interface and power cone test

bodono:
This is the last part of the transition to the new SCS API, this is simply adding the power cone capability and a new test.

Note that the MathProgBase tests start failing randomly with this change, but I'm pretty sure the root cause is not this change but actually this issue: https://github.com/JuliaOpt/SCS.jl/issues/31.


mlubin:
On 0.4, there's a real error here:
```
ERROR: LoadError: LoadError: MethodError: `convert` has no method matching convert(::Type{Ptr{Float64}}, ::Array{Float64,1})
This may have arisen from a call to the constructor Ptr{Float64}(...),
since type constructors fall back to convert methods.
Closest candidates are:
  call{T}(::Type{T}, ::Any)
  convert{T}(::Type{Ptr{T}}, !Matched::UInt64)
  convert{T}(::Type{Ptr{T}}, !Matched::Int64)
```

bodono:
Confirmed this fixes https://github.com/JuliaOpt/SCS.jl/issues/31. It passes all the same tests as master (and fails the same one, which I think should be resolved by https://github.com/staticfloat/homebrew-juliadeps/pull/56/files), so should be safe to merge.

madeleineudell:
This is only failing on the nightly Mac build, passing on all others. I'm going to merge.

madeleineudell:
closes #31

PR #35, opened 2015-06-27
error out when the wrong version of SCS is detected

mlubin:


mlubin:
@tkelman, is this reasonable?

tkelman:
The windows binaries don't get downloaded to `usr`, they get downloaded to `bin32` and `bin64`

mlubin:
I've tweaked the message.

tkelman:
that looks better, though won't deleting the `.gitignore` make the package dirty and never upgrade again?

mlubin:
Tweaked again

mlubin:
Going to bump, looks like any existing installations on Linux are broken now. There's the small issue of ``scs_version`` not being implemented on the previous version though...

bodono:
`scs_version` should be in the compiled library, since its independent of the SCS.jl code.

tkelman:
what was the oldest version of the library that had `scs_version` present?

mlubin:
Previous installs were on ``1.0.7``.

bodono:
It's been in since `1.1.0`. Version `1.0.7` doesn't work with this version of SCS.jl anyway due to the API changes (SDPs etc.), it will error out at runtime.

bodono:
Release https://github.com/cvxgrp/scs/releases/tag/v1.1.0

mlubin:
Yes, it will error out at runtime, but this is the current message:
```
eps tolerance must be positive
ERROR: Validation returned failure
ERROR: NULL input
```
https://groups.google.com/forum/#!topic/julia-opt/FQC3nTAyNvI

bodono:
So, it looks like both issues are caused by the mismatch from `1.0.7` with different versions of SCS.jl. First, when using SCS `1.0.7` but a recent version of SCS.jl the memory is not mapped correctly, which gives the eps tolerance error (other failures in validation are to be expected too, possibly seg faults), then when updating SCS.jl but still using SCS `1.0.7` you get the ccall error, which is more obviously a version mismatch and a "cleaner" error.

mlubin:
We should be able to do a bit better than that, maybe a ``try/catch``?

mlubin:
Going ahead and tagging: https://github.com/JuliaLang/METADATA.jl/pull/2818

src/SCS.jl
```diff
@@ -10,9 +10,14 @@ if VERSION >= v"0.4.0-dev+3844"
     import Base.Libdl: RTLD_LAZY, RTLD_DEEPBIND, RTLD_GLOBAL, dlopen
 end
 
+include("version.jl")
+
 function __init__()
     # default binaries need access to Julia's lapack symbols
     @unix_only dlopen(Base.liblapack_name, RTLD_LAZY|RTLD_DEEPBIND|RTLD_GLOBAL)
+    if SCS_version() != scs_version
+        error("Current SCS version installed is $(SCS_version()), but we require version $scs_version. On Linux and Windows, delete the contents of the `~/.julia/v0.3/SCS/deps` directory except for the files `build.jl` and `.gitignore`, then rerun Pkg.build(\"SCS\"). On OS X, run `using Homebrew; Homebrew.update()` in Julia.")
```
tkelman: this should probably use `@__FILE__` rather than hardcoding `~/.julia/v0.3`

PR #38, opened 2015-07-26
fix deprecation warning in deps/build.jl

tkelman:
slightly roundabout way of doing this that doesn't need Compat


mlubin:
Thanks!

PR #44, opened 2015-09-30
Fix warnings due to deprecated use of 'int()' instead of 'round(Int,)'

marcusps:
Nothing else to it.

### DIFF
--- a/src/SCSSolverInterface.jl
+++ b/src/SCSSolverInterface.jl
@@ -449,7 +449,7 @@ function rescaleconicproblem!(model::SCSMathProgModel)
     SDPstartidx = model.f + model.l + sum(model.q)
     for iSDP = 1:model.ssize
         nSDP = model.s[iSDP]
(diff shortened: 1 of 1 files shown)


src/SCSSolverInterface.jl
```diff
@@ -449,7 +449,7 @@ function rescaleconicproblem!(model::SCSMathProgModel)
     SDPstartidx = model.f + model.l + sum(model.q)
     for iSDP = 1:model.ssize
         nSDP = model.s[iSDP]
-        sSDP = int(nSDP*(nSDP+1)/2)
+        sSDP = round(Int,nSDP*(nSDP+1)/2)
```
mlubin: This should also be ``div(nSDP*(nSDP+1),2)``

marcusps: Will do.

mlubin:
Thanks!

PR #48, opened 2015-11-26
updates for MPB 0.4

mlubin:
Ref JuliaOpt/MathProgBase.jl#91
Don't merge until new MPB is about to be released.

### DIFF
--- a/src/SCS.jl
+++ b/src/SCS.jl
@@ -1,4 +1,4 @@
-isdefined(Base, :__precompile__) && __precompile__()
+__precompile__()
 
 module SCS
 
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -1,7 +1,7 @@
 tests = ["direct.jl",
          "mpb_linear.jl",
-         "options.jl"]
-
+#         "options.jl"]
```
tkelman: huh?

mlubin: Good catch, that was meant to be temporary

PR #55, opened 2016-06-12
fix build on julia 0.5

mlubin:
Ignoring the other deprecation warnings for now, just want to get the build working first.

CC @tkelman


deps/build.jl
```diff
@@ -2,7 +2,14 @@ using BinDeps
...
-if (@osx? (Base.blas_vendor() == :openblas64) : false)
+if VERSION >= v"0.5.0-dev+4679"
+    blasvendor = Base.BLAS.vendor()
+else
+    blasvendor = Base.blas_vendor()
+end
+
+
+if (@osx? (blasvendor == :openblas64) : false)
     aliases = ["libscsdir64"]
```
tkelman: can you remember why this is mac only? lgtm otherwise

mlubin: Not really sure, though it's not easy for me to test now.

PR #60, opened 2016-10-14
Fix depwarn in init function

ranjanan:
`@unix_only` to `if is_unix()`


ranjanan:
@mlubin Could you tag a new release please?

PR #62, opened 2017-01-06
building against mkl with 64int

abulak:


abulak:
the question now is do we care about 32 users of mkl and if we do, how to extract that information from within julia

mlubin:
Julia itself must be aware of the integer size of MKL, so I'm sure there's a way to extract that information at runtime.

abulak:
`versioninfo()` i guess ;-)

abulak:
`Base.USE_BLAS64` is set at the build time, so probably this is the thing to check

abulak:
compiled and tested

deps/build.jl
```diff
@@ -52,6 +52,17 @@ cflags = "-DCOPYAMATRIX -DDLONG -DLAPACK_LIB_FOUND -DCTRLC=1"
 if blasvendor == :openblas64
     cflags = "$cflags -DBLAS64 -DBLASSUFFIX=_64_"
 end
+if blasvendor == :mkl
+    if Base.USE_BLAS64
+        cflags = "$cflags -DMKL_ILP64 -m64 -DBLAS64"
+        ldflags = "$ldflags -lmkl_intel_ilp64"
+    else
+        cflags = "$cflags -m32"
```
tkelman: did you test this? `-m32` is for 32 bit binaries, not integer size

kalmarek: only on 64bit arch so `else` is effectively not tested

PR #66, opened 2017-02-14
update status_mapt to latest constants

kalmarek:
Does `SCS_SOLVED_INACCURATE` qualify for `:UserLimit`?

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -80,13 +80,16 @@ immutable SCSCone
 end
 
 
-# TODO needs to be updated for newest constants
 const status_map = Dict{Int, Symbol}(
+    2 => :UserLimit,
     1 => :Optimal,
-    -2 => :Infeasible,
     -1 => :Unbounded,
+    -2 => :Infeasible,
     -3 => :Indeterminate,
-    -4 => :Error
+    -4 => :Error,
+    -5 => :Interrupted,
+    -6 => :UnboundedInaccurate,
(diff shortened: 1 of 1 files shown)


mlubin:
What are the conditions that could trigger that status? If it's an iteration limit, then yes, I'd say ``UserLimit``.

By the way, we're thinking about reworking the MathProgBase status system because it's pretty broken at the moment, see the discussion at https://github.com/JuliaOpt/MathProgBase.jl/issues/156.

kalmarek:
Yest this can be an iteration limit

mlubin:
``UserLimit`` isn't appropriate if the solver stops because of an iteration limit which the user didn't set.

kalmarek:
I'm pretty sure this is me being dumb, but I have no idea how to distinguish `SCSSolver()` and `SCSSolver(iterations=20000)` in code; for the first hitting its default 20000 iterations shouldn't return `:UserLimit`, while the second should not (then what should it return? `:Inaccurate`?)

How is discussion on extending the number of solve states progressing?

mlubin:
User-provided options are stored in ``SCSMathProgModel.options``, so you can check if ``max_iters`` was set. If SCS stops because it hit the user-provided iteration limit, it should return ``:UserLimit``. If it stops because it hits its internal limit without solving the problem, it should return something like ``:Error``.

I wouldn't expect a revision of the MPB solver statuses until June.

PR #67, opened 2017-02-26
add support for SCSv1.2 on linux

kalmarek:


src/SCS.jl
```diff
@@ -16,7 +16,12 @@ function __init__()
     if is_unix()
         dlopen(Base.liblapack_name, RTLD_LAZY|RTLD_DEEPBIND|RTLD_GLOBAL)
     end
-    if !(vnum.major == 1 && vnum.minor == 1)
+    if is_linux()
+        if !(vnum.major == 1 && vnum.minor == 2)
```
mlubin: If the Julia code is compatible, there's no need to special case ``is_linux`` here. There should be one condition that tests major version == 1 and minor version == 1 or 2.

deps/build.jl
```diff
@@ -19,6 +19,9 @@ if is_apple()
 end
 
 version = "1.1.8"
+if is_linux()
```
mlubin: Condition not needed. Just change ``version`` above.

src/SCS.jl
kalmarek: in this way users will end up with newer version of `SCS.jl` without being forced to upgrade the binary `libscsdir.so`

mlubin: So? They shouldn't be *forced* if the package will work fine. This code here is to avoid segfaults with mismatching binary interfaces. It could be worth printing a warning though.

mlubin:
OS X users get binaries through Homebrew.jl. It's very rare for them to build from source, so this setting does not affect OS X.

You can keep committing to this branch.

src/SCS.jl
kalmarek: should the warning be linux or unix specific?

mlubin: linux specific. We shouldn't print warnings on OS X.

mlubin:
Thanks!

PR #68, opened 2017-03-02
don't mix blas integer size and binary word size

tkelman:
`-m32` and `-m64` are for getting different architecture binaries from a multilib compiler
Julia can use either 32 or 64 bit integers in blas on 64 bit, these aren't coupled


mlubin:
@kalmarek

kalmarek:
Sorry, for that. I used flags from Intel MKL Link Advisor, but without much understanding

PR #70, opened 2017-03-06
build openmp libraries

kalmarek:
things run here and past tests; 
but let someone with more experience with C have a look;

I noticed that multi-threading is disabled if I start julia in parallel, like `julia -p4`
(Guess this is to disable over-threading i.e. multiple processes run multiple threads).

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ println(getvalue(take))
 # [Silver] = 0.4666851698368782
 ```
 
-### Note on Multithreading
-On Linux, SCS.jl will build the SCS libraries with the support of OpenMP, i.e. multithreading. To allow usage of multiple threads set `OMP_NUM_THREADS` env variable to the desired number (roughly the number of physical cores).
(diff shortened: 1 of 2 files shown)


mlubin:
CC @tkelman

mlubin:
Please add a discussion in the README about how to control SCS's use of threads.

tkelman:
do all mac source toolchains have working openmp by now?

kalmarek:
I can't find a reference why the openmp threading is disabled when called from worker;
also https://github.com/JuliaLang/julia/issues/6293 suggests that the number should not exceed the number of _physical cores_ in processor

mlubin:
@kalmarek, have you tested this with both openblas and mkl?

kalmarek:
I did, but to avoid a repetition, let me do it again on a fresh install (tonight or tomorrow)

kalmarek:
this has been tested with both openblas and mkl;
Both openblas- and mkl-linked `libscsdir.so` are able to use up to `$physical_cores`-threads

kalmarek:
I think we can scrap this pull: surprisingly the library is as multithreaded as BLAS is, regardless of linking;

Maybe the part about interaction of `addprocs` and `BLAS.set_num_threads()` in README is worth keeping.

kalmarek:
After the discussion https://github.com/cvxgrp/scs/issues/81 I think I finally got it right;

Linking to openmp is relevant only for Lin-sys phase of the solver, and only if the indirect solver is used;
(this makes linking of `SCSdir` with `-lgomp` superfluous, but the code is much simpler).

PR #75, opened 2017-04-22
setbvec

mlubin:
https://github.com/JuliaOpt/Pajarito.jl/issues/372
We have no tests for this in MPB.
@chriscoey


chriscoey:
@mlubin things are changing in mpb anyway. can we merge so i can use this in pajarito?

PR #78, opened 2017-04-23
Indirect

kalmarek:
Allow the usage of direct and indirect linear solver in SCS;

The default is the direct solver, as before;
To explicitly select solver use 'linearsolver' keyword in SCSSolver constructor;
Possible values: `SCS.Direct`, `SCS.Indirect`;


mlubin:
The OS X and windows binaries will need to be updated to include the indirect library before tests will pass.

nantonel:
@mlubin @kalmarek hi all. I'm wondering if the binary have been updated and this PR can be merged?

mlubin:
The PR has merge conflicts, so it cannot be merged. If it's updated, I'd be willing to merge without resolving the binary issue because `master` doesn't have binaries working yet anyway.

kalmarek:
I'm in the process of rebasing it on current master;
It will probably take me another 2-3 days.

PR #84, opened 2017-11-26
Remove deprecate size argument for cones

blegat:
This PR deprecates the arguments `qsize`, `ssize` and `psize`. These indicate the length of the vectors `q`, `s` and `p` in C but they are not necessary in Julia as the length can be queried from the vector.
Having a separate length is therefore redundant and could potentially create bugs very difficult to debug for the user.
I have create methods without the argument and in old methods, I have added a deprecation warning with `@assert` to check that the size argument is indeed the size of the vector. This is in case a user have a longer vector and actually needs the rest of the vector to be ignore and therefore depends on the feature. This user will get an assert error which indicates he is doing something wrong.


mlubin:
Ok to merge if travis-ci/pr is green on linux.

blegat:
I have rebased the PR and Linux is passing now on v0.6 and nightly :)

PR #89, opened 2017-11-30
Bump to v2.0.2

bodono:
Now working on windows again. Update the tags to the latest release.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -19,7 +19,7 @@ if is_apple()
 end
 
 version = "2.0.2"
-win_version = "2.0.2"
+win_version = "2.0.2" # The windows binaries are not consistent with this version yet.
 
 provides(Sources, URI("https://github.com/cvxgrp/scs/archive/v$version.tar.gz"),
     [scs], os=:Unix, unpacked_dir="scs-$version")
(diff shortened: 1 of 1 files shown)


mlubin:
So what type is used for integers now? If it's `long`, then we should switch Julia's `Int` to `Clong` in the appropriate places for cross-platform compatibility.

deps/build.jl
```diff
@@ -19,7 +19,7 @@ if is_apple()
 end
 
 version = "2.0.2"
-win_version = "2.0.2" # 2.0.* is not yet working on windows
+win_version = "2.0.2"
```
mlubin: Why remove this comment when the current build script still does not support windows?

mlubin:
I see that we switched from `Clong` to `Int` some time ago... https://github.com/JuliaOpt/SCS.jl/commit/51cc9f0061cee0d0a682c191af7ab06e1d977b9c. Not sure why.

deps/build.jl
bodono: SCS 2.0.2 works fine with windows, the binaries need to be updated for SCS.jl. I updated the comment.

mlubin: Ok, good to clarify.

PR #93, opened 2017-12-08
reduce alpha to 1.5

bodono:
Alpha of 1.5 has been the default in SCS for a while, because for some problems 1.8 can be too aggressive. See the discussion here: https://github.com/JuliaOpt/MathOptInterfaceSCS.jl/issues/1#issuecomment-350217337

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -42,7 +42,7 @@ struct SCSSettings
     rho_x::Cdouble # x equality constraint scaling: 1e-3
     max_iters::Int # maximum iterations to take: 5000
     eps::Cdouble # convergence tolerance: 1e-5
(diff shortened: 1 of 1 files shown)


bodono:
The other interfaces to SCS use a helper function [here](https://github.com/cvxgrp/scs/blob/f81d51aba3408b07062da8f719d412a4426f960d/src/util.c#L202) which just sets the default parameters, we might want to consider doing the same for SCS.jl.

mlubin:
+1 for using `set_default_settings`, but I'm also confused by https://github.com/JuliaOpt/MathOptInterfaceSCS.jl/issues/1#issuecomment-350256684.

PR #95, opened 2017-12-08
Indirect take two

kalmarek:
Please test for `os=:Darwin` and `os=:Windows`.

Note that `using SCS` will fail if You have previous version installed with
```julia
LoadError: UndefVarError: direct not defined
```
This is because of precompilation and code generation in `c_wrapper.jl` I don't know how to get around it to tell user that running `Pkg.build("SCS")` is required.


deps/build.jl
```diff
@@ -4,25 +4,25 @@ using BinDeps
 
 blasvendor = Base.BLAS.vendor()
 
+libs = Dict(:direct=>"libscsdir", :indirect=>"libscsindir")
```
mlubin: These names aren't going to change frequently, so just remove this extra level of indirection.

deps/build.jl
```diff
@@ -4,25 +4,25 @@ using BinDeps
...
 end
 
-scs = library_dependency("scs", aliases=aliases)
+direct = library_dependency("direct", aliases=[libs[:direct]])
+indirect = library_dependency("indirect", aliases=[libs[:indirect]])
 
 if is_apple()
     using Homebrew
-    provides(Homebrew.HB, "scs", scs, os = :Darwin)
+    provides(Homebrew.HB, [libs[:direct], libs[:indirect]], [direct, indirect], os = :Darwin)
```
mlubin: Revert this change. Homebrew will still provide a single package called "scs".

deps/build.jl
```diff
@@ -31,13 +31,13 @@ provides(Sources, URI("https://github.com/cvxgrp/scs/archive/v$version.tar.gz"),
 # CFLAGS="-DDLONG -DCOPYAMATRIX -DUSE_LAPACK -DCTRLC=1" LDFLAGS="-L$HOME/julia32/usr/bin -lopenblas" make CC=i686-w64-mingw32-gcc out/libscsdir.dll
 # mv out bin32
 provides(Binaries, URI("https://cache.julialang.org/https://bintray.com/artifact/download/tkelman/generic/scs-$win_version-r2.7z"),
-    [scs], unpacked_dir="bin$(Sys.WORD_SIZE)", os = :Windows,
+    [direct, indirect], unpacked_dir="bin$(Sys.WORD_SIZE)", os = :Windows,
```
mlubin: Just comment out this "provides" for windows and leave a TODO. That's better than leaving code that is known to not work.

src/c_wrapper.jl
```diff
@@ -44,6 +44,13 @@ function SCS_solve(m::Int, n::Int, A::SCSVecOrMatOrSparse, b::Array{Float64},
         slack::Vector{Float64}=Float64[];
         options...)
 
+    T = SCS.Indirect # the default method
```
mlubin: At the level of the C wrapper, `Direct` and `Indirect` should be parameters to `SCS_solve` as below, not stuffed into `options`.

mlubin:
> I don't know how to get around it to tell user that running Pkg.build("SCS") is required.

That's fine, because nobody will have SCS 2.0 installed anyway, and they'll be forced to rebuild.

deps/build.jl
kalmarek: Do You mean
```julia
provides(Homebrew.HB, "scs", [direct, indirect], os = :Darwin)
```
so that Homebrew "scs" package will provide both direct and indirect lib?

mlubin: Yes. But probably best to comment out this code also and leave a TODO because it doesn't work.

src/c_wrapper.jl
kalmarek: I moved this default and options "cleaning" to `optimize!` in `SCSSolverInterface.jl`

deps/build.jl
```diff
@@ -64,13 +59,21 @@ ENV2["CFLAGS"] = cflags
...
+        GetSources(direct)
         CreateDirectory(joinpath(prefix, "lib"))
-        FileRule(joinpath(prefix, "lib", libname), @build_steps begin
+
+        FileRule(joinpath(prefix, "lib", "libscsdir.$(Libdl.dlext)"), @build_steps begin
             ChangeDirectory(srcdir)
-            setenv(`make BLASLDFLAGS= out/$libname`, ENV2)
-            `mv out/$libname $prefix/lib`
+            setenv(`make BLASLDFLAGS= out/libscsdir.$(Libdl.dlext)`, ENV2)
+            `mv out/libscsdir.so $prefix/lib`
```
mlubin: The build is failing on OS X because of the `.so` extension here: https://travis-ci.org/JuliaOpt/SCS.jl/jobs/313571982#L360
Please try to get OS X passing.

mlubin:
Thanks!

PR #100, opened 2018-05-04
Julia 0.7

kalmarek:
this is an almost automatic port to `julia-0.7`; The only notable change is how `kwargs` are handled (as `NamedTuple`)


deps/build.jl
```diff
@@ -2,13 +2,16 @@ using BinDeps
 
 @BinDeps.setup
 
-blasvendor = Base.BLAS.vendor()
+using Libdl
```
blegat: You may need Compat for Libdl, LinearAlgebra, and Sys

kalmarek: Yep, but to be honest I don't know what to do with `NamedTuples`, as on `0.6` we get
```julia
julia> (x=1,)
ERROR: syntax: assignment not allowed inside tuple
```

kalmarek:
after reading the very long discussion in https://github.com/JuliaLang/julia/pull/22194 I am not sure if NamedTuples will be ever supported by Compat...

Any idea how to support both 0.6 and 0.7?

mlubin:
What's the situation where you need to use NamedTuples? We haven't had an issue with this for other solvers.

mlubin:
You can define an `addoption!` function, e.g.,
```julia
if VERSION >= v"0.7-"
    addoption!(options, name, value) = ...
else
    addoption!(options, name, value) = ...
end
```
If the 0.7 version doesn't parse on 0.6, you can wrap it in `eval(parse())`.

ECOS is passing on nightly travis.

src/types.jl
```diff
@@ -68,7 +68,7 @@ struct SCSData
     stgs::Ptr{SCSSettings}
 end
 
-struct SCSSolution
+@compat struct SCSSolution
```
mlubin: This `@compat` shouldn't be needed.

mlubin:
LGTM, thanks!

PR #103, opened 2018-07-23
support blas-reference

kalmarek:
for those brave to build julia on their own

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -56,6 +56,9 @@ if blasvendor == :mkl
     cflags = "$cflags -fopenmp"
     ldflags = "$ldflags -lmkl_gnu_thread -lmkl_rt -lmkl_core -lgomp"
 end
+if blasvendor == :unknown
+    ldflags = "$ldflags -lblas -llapack"
+end
 
 ENV2 = copy(ENV)
 ENV2["LDFLAGS"] = ldflags


mlubin:
Could you add a comment describing in what situations `blasvendor == :unknown` can occur?

kalmarek:
e.g. when julia is compiled with `blas-reference`. This may happen e.g. in certain gentoo Linux configurations

mlubin:
I meant a comment in the code :)

kalmarek:
ok, so it turns out that it is not that simple (i.e. we could scrap this pull)

julia comes with its own (internal) BLAS library, and since we don't link `libscs*` with any blas/lapack, the library uses those symbols which are loaded with julia, i.e. julia's BLAS. Now if julia is to run on a hardened system (this was the actual cause fo problems) linking must be immediate, hence the needed `-lblas -llapack`. However then we link against whatever system provides we risk using two BLASes at the same time.

A solution would be to find out the exact path of BLAS julia uses (internal, or systems) and link against it, but I have no idea how to do it.

mlubin:
I would recommend looking at BinaryBuilder before putting too much more effort into this build script (ideally the BinaryBuilder version will replace this). BinaryBuilder is designed to link against the libraries that ship with Julia (including gfortran and BLAS). I don't know, however, what this means for MKL builds.

kalmarek:
BinaryBuilder was more pleasant than expected:

https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/313

There is a problem with building for windows with openmp enabled. Everything else works fine.

At least in the wizard ;-)

mlubin:
Glad to hear that you got it almost working :)

Yes, there are lots of trade-offs with BinaryBuilder. We should ensure that users still have an option to build from source, and it's also not necessary to immediately get rid of the BinDeps script for linux (we still use it for Ipopt, for example). But for MacOS and Windows binaries I don't know of any other practical way that we can move forward and get something workable enough to tag a new release.

PR #104, opened 2018-08-08
Update to MOI v0.5

blegat:
and replace the appveyor script with [the new one](https://discourse.julialang.org/t/ann-improved-appveyor-scripts/12754)


mlubin:
This is a bit old already, but I'll merge so that we can move onto MOI 0.6.

PR #105, opened 2018-09-14
Update to MOI v0.6

blegat:


blegat:
Windows is already failing on master and not due to this PR

PR #107, opened 2018-09-19
BB try 2

juan-pablo-vielma:


blegat:
Looks like it's green on Travis for Julia v0.6 to Julia v1.0 :tada: 

On Arch Linux with [Julia v1.0 from pacman](https://www.archlinux.org/packages/community/x86_64/julia/) though:
```
$ ldd libscsindir.so 
	linux-vdso.so.1 (0x00007ffcf30df000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f20add1f000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007f20add15000)
	libopenblas64_.so.0 => not found
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f20adb51000)
	/usr/lib64/ld-linux-x86-64.so.2 (0x00007f20ae0fa000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f20adb30000)
$ ldd libscsdir.so 
	linux-vdso.so.1 (0x00007ffe9a3a6000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007fe75ef93000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007fe75ef89000)
	libopenblas64_.so.0 => not found
	libc.so.6 => /usr/lib/libc.so.6 (0x00007fe75edc5000)
	/usr/lib64/ld-linux-x86-64.so.2 (0x00007fe75f37a000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fe75eda4000)
$ ls -a /usr/lib/julia 
.          libcamd.so             libccolamd.so  libdSFMT.so  libmpfr.so        libopenblas.so    libopenlibm.so.2    libpcre2-8.so.0      libsuitesparseconfig.so    sys.so
..         libccalltest.so        libcholmod.so  libgit2.so   libmpfr.so.6      libopenblas.so.0  libopenlibm.so.2.5  libpcre2-8.so.0.6.0  libsuitesparse_wrapper.so
libamd.so  libccalltest.so.debug  libcolamd.so   libgmp.so    libmpfr.so.6.0.1  libopenlibm.so    libpcre2-8.so       libspqr.so           libumfpack.so

$ julia
julia> using LinearAlgebra
julia> LinearAlgebra.BLAS.vendor()
:openblas
```
If I am not mistaken, the fact that `LinearAlgebra.BLAS.vendor()` is `:openblas` and not `:openblas64` means that the library will be called `libopenblas.so` and not `libopenblas64_.so`.

kalmarek:
also please provide a way for building scs libs directly

juan-pablo-vielma:
Do the tests pass in Arch Linux?

On Sep 20, 2018, at 5:39 AM, Benoît Legat <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks like it's green on Travis for Julia v0.6 to Julia v1.0 🎉

On Arch Linux with Julia v1.0 from pacman<https://www.archlinux.org/packages/community/x86_64/julia/> though:

$ ldd libscsindir.so
        linux-vdso.so.1 (0x00007ffcf30df000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f20add1f000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007f20add15000)
        libopenblas64_.so.0 => not found
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f20adb51000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007f20ae0fa000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f20adb30000)
$ ldd libscsdir.so
        linux-vdso.so.1 (0x00007ffe9a3a6000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007fe75ef93000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007fe75ef89000)
        libopenblas64_.so.0 => not found
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fe75edc5000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007fe75f37a000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fe75eda4000)
$ ls -a /usr/lib/julia
.          libcamd.so             libccolamd.so  libdSFMT.so  libmpfr.so        libopenblas.so    libopenlibm.so.2    libpcre2-8.so.0      libsuitesparseconfig.so    sys.so
..         libccalltest.so        libcholmod.so  libgit2.so   libmpfr.so.6      libopenblas.so.0  libopenlibm.so.2.5  libpcre2-8.so.0.6.0  libsuitesparse_wrapper.so
libamd.so  libccalltest.so.debug  libcolamd.so   libgmp.so    libmpfr.so.6.0.1  libopenlibm.so    libpcre2-8.so       libspqr.so           libumfpack.so

$ julia
julia> using LinearAlgebra
julia> LinearAlgebra.BLAS.vendor()
:openblas


If I am not mistaken, the fact that LinearAlgebra.BLAS.vendor() is :openblas and not :openblas64 means that the library will be called libopenblas.so and not libopenblas64_.so.

mlubin:
BinaryBuilder is not designed to support Julia versions outside the official binaries. We don't have the ability to make this work on our own (without changes to BinaryBuilder). What we can do is provide an environment variable that lets you override the library search path if you choose to compile SCS manually (e.g., like you can install Ipopt manually).

Another option would be to leave the BinDeps script for Linux and use BinaryProvider on Mac and Windows. This is painful to maintain though, and BinDeps might go away soon.

kalmarek:
@blegat Arch ships a non-standard build which is linked against systems openblas

juan-pablo-vielma:
I will add the JULIA_SCS_LIBRARY_PATH support for custom builds. That is relatively low cost for the user and should fix almost any situation.

blegat:
What is tricky here is that the binaries is compiled with a 64_ suffix to all blas calls so linking it with the system openblas will not work. I have tried renaming the library names with the 64_ suffix. This fools BinaryProvider but when running the tests, I get
```
julia: symbol lookup error: /home/blegat/.julia/dev/SCS/deps/usr/lib/libscsdir.so: undefined symbol: dgemv_64_
```
since the symbol in the library is `dgemv_` and not `dgemv_64_`.

juan-pablo-vielma:
@blegat Correct, in fact for Mac OS I had to rename the library so it would be compatible with Julia's naming convention (there is a bug in the OpenBLASBuilder code that was supposed to add this so I will open an issue):
https://github.com/JuliaOpt/SCSBuilder/blob/9754b1b364ec257410faf8db6c1ff8f4857cc7d4/build_tarballs.jl#L28-L30

I believe the plan for BB to resolve this is for BB to "learn" what libraries (with specific versions) are required by a binary dependency and get BP to install tested versions without redundancy (e.g. with our COIN-OR solution we install multiple copies of CoinOsi, which I need to double check does not cause any problems). These would still not be the "system" libraries, but the same way we can let the user override BP for the main library (e.g. SCS), we should be able to let the user override the BP provided dependencies for that library (e.g. BLAS or GOMP).

blegat:
Would it be possible to binaries with `openblas` and with `openblas64_` or is BB restricted to one binary by platform ?
In
https://github.com/JuliaOpt/SCSBuilder/blob/9754b1b364ec257410faf8db6c1ff8f4857cc7d4/build_tarballs.jl#L32
we just need to tweak `BLAS64=1 BLASSUFFIX=_64_` and `-lopenblas64_` to use `openblas` instead of `openblas64_` like what is done in
https://github.com/JuliaOpt/SCS.jl/blob/8cf4a241df60373dc6540d3d0d691e068b9cfa93/deps/build.jl#L46-L48
The system openblas can be used to precompile for `openblas`

juan-pablo-vielma:
Just thought about something else. We could have a JULIA_XXX_CUSTOM_BUILDER variable with an url to an alternative `build.jl` script. We could have some those builders in JuliaOpt and label them as "experimental" or "advanced" and then we do not need to be that careful on their testing.

blegat:
I like the idea of community driven alternative builders.
In a short term perspective, I think a warning in case `Libdl.dlopen_e("libopenblas64_")` returns NULL is sufficient.

juan-pablo-vielma:
Ok, added custom install (tested in linux) with instructions and a warning about non-official binaries. @blegat what does the error in Arch Linux say? is this warning enough? Just checking for `Libdl.dlopen_e("libopenblas64_")` would not work because in 32bit architectures it would fail on the official Julia binaries.

blegat:
I just tried the README instructions and it works fine until I run the tests and get
```
m and n must both be greater than 0; m = 1, n = 0
ERROR: Validation returned failure

signal (11): Segmentation fault
```
Then I did `make clean && make DLONG=1`, run the tests again and it works fine.
Maybe we should add in the README to compile with `DLONG=1` if compiling from source.

deps/build.jl
```diff
@@ -1,83 +1,54 @@
...
+
+custom_library = false
+if haskey(ENV,"JULIA_SCS_LIBRARY_PATH")
+    custom_products = [LibraryProduct(ENV["JULIA_SCS_LIBRARY_PATH"],product.libnames,product.variable_name) for product in products]
+    if all(satisfied(p; verbose=verbose) for p in custom_products)
+        products = custom_products
+        custom_library = true
     else
-        ldflags = "$ldflags -lmkl_intel"
+        warn("Could not install custom libraries from $(ENV["JULIA_SCS_LIBRARY_PATH"]). Falling back to BinaryProvider")
```
mlubin: `warn` is not defined in Julia 1.0. Use `Compat.@warn` instead.

juan-pablo-vielma:
looks like the custom build is not working on 0.7/1.0. I am checking now.

juan-pablo-vielma:
ok, looks like warnings don't show up on the REPL in Pkg3. I will ask around and if not switch it to an informative error

juan-pablo-vielma:
I convinced myself that it should be an error instead of a warning. The error explains how to fall back to BP.

juan-pablo-vielma:
BTW @kalmarek the final `build_tarballs.jl` is almost the same as yours, so it was mostly waiting for BB to stabilize :-) This is noted in https://github.com/JuliaOpt/SCSBuilder/blob/master/build_tarballs.jl

PR #110, opened 2018-09-26
Enh/options

kalmarek:
Would be great if someone could test it on the MOI side;

I took a simple route of extracting the `Direct`/`Indirect` type from `options` in `optimize!` (as it was done in MPB), but that could be easily changed to `Optimizer` (and `SCSMathProgModel`) parametrized by the type

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -60,9 +60,11 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     maxsense::Bool
     data::Union{Nothing, ModelData} # only non-Void between MOI.copy_to and MOI.optimize!
     sol::MOISolution
-    function Optimizer()
-        new(ConeData(), false, nothing, MOISolution())
+    options
+    function Optimizer(; options...)
+        new(ConeData(), false, nothing, MOISolution(), options)
     end
+
 end
 
 function MOI.is_empty(optimizer::Optimizer)
@@ -282,7 +284,15 @@ function MOI.optimize!(optimizer::Optimizer)
     objconstant = optimizer.data.objconstant
     c = optimizer.data.c
     optimizer.data = nothing # Allows GC to free optimizer.data before A is loaded to SCS
-    sol = SCS_solve(SCS.Indirect, m, n, A, b, c, cone.f, cone.l, cone.qa, cone.sa, cone.ep, cone.ed, cone.p)
+
+    T = SCS.Indirect # the default method
+    opts = Dict(optimizer.options)
+    if :linear_solver in keys(opts)
+        T = opts[:linear_solver]
+    end
+    options = [(k,v) for (k,v) in optimizer.options if k !=:linear_solver]
+
+    sol = SCS_solve(T, m, n, A, b, c, cone.f, cone.l, cone.qa, cone.sa, cone.ep, cone.ed, cone.p, options...)
     ret_val = sol.ret_val
     primal = sol.x
     dual = sol.y
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -58,6 +58,11 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
         k in all_kwargs || throw(ArgumentError("Unknown keyword passed to SCS solver: $k"))
     end
 
+    for param in keys(opts)
+        # when dropping support for 0.6 could be replaced with setproperty!
+        setfield!(settings[], param, convert(fieldtype(SCSSettings, param), opts[param]))
+    end
+
     cone = Ref(SCSCone(f, l, q, s, ep, ed, p))
     info = Ref(SCSInfo())
 
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -51,6 +51,13 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
     data = Ref(SCSData(m, n, Base.unsafe_convert(Ptr{SCSMatrix}, matrix), pointer(b), pointer(c), Base.unsafe_convert(Ptr{SCSSettings},settings)))
     SCS_set_default_settings!(T, data)
 
+    opts = Dict(options)
+
+    all_kwargs = [:scale, :rho_x, :max_iters, :eps, :alpha, :cg_rate, :acceleration_lookback, :normalize, :verbose, :warm_start]
+    for k in keys(opts)
+        k in all_kwargs || throw(ArgumentError("Unknown keyword passed to SCS solver: $k"))
+    end
+
     cone = Ref(SCSCone(f, l, q, s, ep, ed, p))
     info = Ref(SCSInfo())
 
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -47,8 +47,10 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
 
     managed_matrix = ManagedSCSMatrix(m, n, A)
     matrix = Ref(SCSMatrix(managed_matrix))
-    settings = Ref(SCSSettings(;options...))
+    settings = Ref(SCSSettings())
     data = Ref(SCSData(m, n, Base.unsafe_convert(Ptr{SCSMatrix}, matrix), pointer(b), pointer(c), Base.unsafe_convert(Ptr{SCSSettings},settings)))
+    SCS_set_default_settings!(T, data)
+
     cone = Ref(SCSCone(f, l, q, s, ep, ed, p))
     info = Ref(SCSInfo())
 
@@ -86,6 +88,11 @@ end
 # Take Ref{}s because SCS might modify the structs
 for (T, lib) in zip([SCS.Direct, SCS.Indirect], [SCS.direct, SCS.indirect])
     @eval begin
+
+        function SCS_set_default_settings!(::Type{$T}, data::Ref{SCSData})
+            ccall((:scs_set_default_settings, $lib), Nothing, (Ref{SCSData}, ), data)
+        end
+
         function SCS_init(::Type{$T}, data::Ref{SCSData}, cone::Ref{SCSCone}, info::Ref{SCSInfo})
 
             p_work = ccall((:scs_init, $lib), Ptr{Nothing},
--- a/src/types.jl
+++ b/src/types.jl
@@ -37,7 +37,7 @@ SCSMatrix(m::ManagedSCSMatrix) =
     SCSMatrix(pointer(m.values), pointer(m.rowval), pointer(m.colptr), m.m, m.n)
 
 
-struct SCSSettings
+mutable struct SCSSettings
     normalize::Int # boolean, heuristic data rescaling: 1
     scale::Cdouble # if normalized, rescales by this factor: 1
     rho_x::Cdouble # x equality constraint scaling: 1e-3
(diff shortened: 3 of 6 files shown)


kalmarek:
btw. if tests fail on SCS-2.* because of instability You may always set `alpha=1.9` and `acceleration_lookback=1` to bring back the old behaviour

kalmarek:
The whole code would be much cleaner if `SCS(set_default_settings)` wouldn't require the whole `SCSData`, but operate directly on `SCSSettings`. Any particular reason for this choice @bodono ?

src/MOIWrapper.jl
```diff
@@ -282,7 +284,15 @@ function MOI.optimize!(optimizer::Optimizer)
     objconstant = optimizer.data.objconstant
     c = optimizer.data.c
     optimizer.data = nothing # Allows GC to free optimizer.data before A is loaded to SCS
-    sol = SCS_solve(SCS.Indirect, m, n, A, b, c, cone.f, cone.l, cone.qa, cone.sa, cone.ep, cone.ed, cone.p)
+
+    T = SCS.Indirect # the default method
```
mlubin: Nit: Please use a more descriptive variable name than `T`. Maybe `linear_solver_method` or just `linear_solver`.

src/c_wrapper.jl
```diff
@@ -47,8 +47,22 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
+
+    opts = Dict(options)
+
+    all_kwargs = [:scale, :rho_x, :max_iters, :eps, :alpha, :cg_rate, :acceleration_lookback, :normalize, :verbose, :warm_start]
+    for k in keys(opts)
+        k in all_kwargs || throw(ArgumentError("Unknown keyword passed to SCS solver: $k"))
+    end
+
+    for param in keys(opts)
+        # when dropping support for 0.6 could be replaced with setproperty!
```
mlubin: What could be replaced exactly? To make sure this happens, you could use a VERSION if/else.

src/types.jl
```diff
@@ -37,7 +37,7 @@ SCSMatrix(m::ManagedSCSMatrix) =
     SCSMatrix(pointer(m.values), pointer(m.rowval), pointer(m.colptr), m.m, m.n)
 
 
-struct SCSSettings
+mutable struct SCSSettings
```
mlubin: I don't think this change is valid. `SCSSettings` needs to mirror the C struct in order for the `ccall`s to be correct. `mutable struct`s can't be used to mirror C structs AFAIK.

test/MOIWrapper.jl
```diff
@@ -14,17 +14,19 @@ MOIU.@model(SCSModelData,
...
             (MOI.VectorOfVariables,),
             (MOI.VectorAffineFunction,))
-const optimizer = MOIU.CachingOptimizer(SCSModelData{Float64}(), SCS.Optimizer())
+for T in [SCS.Direct, SCS.Indirect]
+    optimizer = MOIU.CachingOptimizer(SCSModelData{Float64}(), SCS.Optimizer(linear_solver=T))
 
-# linear9test needs 1e-3 with SCS < 2.0 and 5e-1 with SCS 2.0
-# linear2test needs 1e-4
-const config = MOIT.TestConfig(atol=1e-4, rtol=1e-4)
+    # linear9test needs 1e-3 with SCS < 2.0 and 5e-1 with SCS 2.0
```
mlubin: Why do we care about tolerances for an SCS version that we don't use anymore?

src/types.jl
```diff
@@ -37,7 +37,7 @@ SCSMatrix(m::ManagedSCSMatrix) =
...
     normalize::Int # boolean, heuristic data rescaling: 1
```
mlubin: Please update these comments to reflect that the default parameters are set by `SCS_set_default_settings!` (i.e., remove the values because we have no way to make sure they are correct).

src/c_wrapper.jl
```diff
@@ -86,6 +100,11 @@ end
 # Take Ref{}s because SCS might modify the structs
 for (T, lib) in zip([SCS.Direct, SCS.Indirect], [SCS.direct, SCS.indirect])
     @eval begin
+
+        function SCS_set_default_settings!(::Type{$T}, data::Ref{SCSData})
```
mlubin: Naming nit: The name implies modification, so JuMP style recommends dropping the `!` from the name (http://www.juliaopt.org/JuMP.jl/latest/style.html#Use-of-!-1). This is also consistent with `SCS_init`, `SCS_solve`, and `SCS_finish`.

test/MOIWrapper.jl
```diff
@@ -14,17 +14,19 @@ MOIU.@model(SCSModelData,
...
+    # linear2test needs 1e-4
+    config = MOIT.TestConfig(atol=1e-4, rtol=1e-4)
 
-@testset "Continuous linear problems" begin
-    # AlmostSuccess for linear9 with SCS 2
-    MOIT.contlineartest(MOIB.SplitInterval{Float64}(optimizer), config, ["linear9"])
-end
+    @testset "Continuous linear problems" begin
+        # AlmostSuccess for linear9 with SCS 2
+        MOIT.contlineartest(MOIB.SplitInterval{Float64}(optimizer), config, ["linear9"])
```
mlubin: Is there a reasonable parameter setting that gets `linear9` passing?

test/MOIWrapper.jl
blegat: The tolerance needed for SCS < 2.0 is mentionned to show that this is a regression

test/MOIWrapper.jl
blegat: We need to set tolerance to 5e-1 last time I tested but it seems a bit restrictive to set it for all tests.
`linear9test(optimizer, MOIT.TestConfig(atol=0.5, rtol=0.5)` could be run separately in the line below

kalmarek:
if user supplies a wrong kwarg it will result in
```julia
SCSSettings(SCS.Indirect, eps=1e-12, epps=0.1)
ERROR: MethodError: no method matching _SCS_user_settings(::SCS.SCSSettings; epps=0.1, eps=1.0e-12)
Closest candidates are:
  _SCS_user_settings(::Any; normalize, scale, rho_x, max_iters, eps, alpha, cg_rate, verbose, warm_start, acceleration_lookback) at /home/kalmar/.julia/v0.6/SCS/src/types.jl:70 got unsupported keyword argument "epps"
....
```
I think it is informative enough

mlubin:
> I think it is informative enough

Not really, see the style discussion [here](http://www.juliaopt.org/JuMP.jl/latest/style.html#User-facing-MethodError-1). Please at least add a TODO to improve the error message.

test/MOIWrapper.jl
mlubin: It's unlikely that this code will handle SCS < 2.0 in the future, so this isn't the right place to discuss regressions.

bodono:
> The whole code would be much cleaner if `SCS(set_default_settings)` wouldn't require the whole `SCSData`, but operate directly on `SCSSettings`. Any particular reason for this choice @bodono ?

I don't think there is a particularly good reason for that choice, and it should probably just take the settings struct, but changing it now would be breaking.

kalmarek:
> > I think it is informative enough
> 
> Not really, see the style discussion [here](http://www.juliaopt.org/JuMP.jl/latest/style.html#User-facing-MethodError-1). Please at least add a TODO to improve the error message.

is `try ... catch MethodError error("You probably supplied unrecognized option to SCS.Optimizer ")` acceptable?

kalmarek:
@bodono It'd be great if it could be changed the next release, but it's definitely not critical

mlubin:
> is try ... catch MethodError error("You probably supplied unrecognized option to SCS.Optimizer ") acceptable?

1. You can't catch specific error types. You have to catch all errors, check the type, and rethrow if it's not what you expected.

2. This is not convincingly better than the `MethodError` you showed because it hides the name of the incorrect option.

kalmarek:
```julia
julia> SCSSettings(SCS.Indirect, epps=0.1, e=0.3, eps=1e-12)
ERROR: ArgumentError: Unrecognized options passed to the SCSSolver: epps, e
Stacktrace:
....
```

src/types.jl
```diff
@@ -38,24 +38,61 @@ SCSMatrix(m::ManagedSCSMatrix) =
...
+    dummy_data = Ref(SCSData(0,0, Base.unsafe_convert(Ptr{SCSMatrix}, matrix),
+        pointer([0.0]), pointer([0.0]),
+        Base.unsafe_convert(Ptr{SCSSettings}, default_settings)))
+    SCS_set_default_settings(linear_solver, dummy_data)
+    try
+        settings = _SCS_user_settings(default_settings[]; options...)
+    catch err
+        if err isa MethodError
+            SCS_kwargs = fieldnames(default_settings[])
+            kwargs = [err.args[1][i*2-1] for i in 1:(length(err.args[1]) ÷ 2)]
```
mlubin: There are a number of ways that this code could stop working unintentionally, so please add tests. The structure of `err` could be changed across Julia versions or `_SCS_user_settings` could be changed.

kalmarek: I can `@test_throws ArgumentError ...` against misspelled options, but that's it. 
`_SCS_user_settings` is an internal function, it won't change beyond our control (btw. why is `SCSSettings` exported? I had always assumed that user is not supposed to construct it?)

I'd use the standard `MethodError` which actually tells exactly which options were not recognized, but that clashes with the style guide. What are Your suggestions on handling this?

blegat: > I can @test_throws ArgumentError ... against misspelled options, but that's it.

That would make good tests :)

> it won't change beyond our control

yes but when we change it we might forget to change `MOIWrapper.jl`.

test/options.jl
```diff
@@ -47,3 +47,10 @@ MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
 MathProgBase.setwarmstart!(m, primal_sol; dual_sol = dual_sol, slack = slack)
 MathProgBase.optimize!(m)
 @test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-10, rtol=0.0)
+
+# tests for incorrect options
+s = SCSSolver(eps=1e-12, epps=1.0)
+m = MathProgBase.ConicModel(s)
+MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
+
+@test_throws ArgumentError MathProgBase.optimize!(m)
```
mlubin: Is there some way to test the text of the error message? That's pretty important in this case.

kalmarek: https://github.com/JuliaLang/julia/blob/e7d15d4a013a43442b75ba4e477382804fa4ac49/stdlib/Test/src/Test.jl#L532
`@test_throws` wants type, not an instance

mlubin: Yes, `@test_throws` is probably not the way to do it. It may need a `try/catch` where you test the error message inside the `catch` and test that the `catch` block executed.

ccoffrin:
Just as a general comment.  There were some fairly big changes in the algorithm's performance since the last version.  The change in the default behavior let my [CI to hang indefinitely](https://travis-ci.org/lanl-ansi/PowerModels.jl/builds/434408934) while waiting for SCS to converge.

In the end adding the parameter `acceleration_lookback=1` resolved the issue.  However, I don't think I would have been able to figure this out without Miles' assistance.

kalmarek:
@ccoffrin  its a known issue: https://github.com/cvxgrp/scs/issues/93
the new (unreleased) version 2.1.0 in https://github.com/cvxgrp/scs/tree/2.1.0 tries to use L1 cost which alleviates the problem, but not fully;

`alpha=1.9, acceleration_lookback=1` reverses SCS to pre-2.0 behaviour

test/MOIWrapper.jl
kalmarek:
```julia
 SCS.Optimizer(linear_solver=T, eps=1e-8))
 config = MOIT.TestConfig(atol=1e-5)
```
passes `MOIT.contlineartest(MOIB.SplitInterval{Float64}(optimizer), config)` with both `Direct` and `Indirect` solver;

btw. You need to add some tests for passing options to `SCS.Optimizer`. only playing with this allowed me to catch missing `;` in the argument list.

ccoffrin: I think I just ran into the same issue where `SCS.Optimizer` does not accept solver options.

kalmarek:
I separated the option checking into `sanatize_SCS_options` function, there is no parsing of exceptions anymore (which was tad fragile);

please have a look again

PR #119, opened 2018-10-25
Add using Compat in tests

blegat:
Otherwise, the tests fails at `https://github.com/JuliaOpt/MathOptInterface.jl/blob/00b5965ea1491a893af01c0bf8b5501a0f11c4bb/src/Utilities/model.jl#L550` with MOI v0.6.2 because `Nothing` is not defined on Julia v0.6.
See https://juliarun-ci.s3.amazonaws.com/a8e4d0c3b1d78372dcdb95f9a6b6c484bf33a851/SCS_with_pull_request_of_MathOptInterface_on_julia_0_6.log

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,4 @@
+using Compat
 using Compat.Test
 using SCS
 


mlubin:
Can this be fixed instead by changing `name_to_var::Union{Dict{String, $VI}, Nothing}` to `name_to_var::Union{Dict{String, $VI}, $MOIU.Nothing}`?

mlubin:
Adding `Compat` hides a hygiene issue in the macro. It could be broken on Julia 1.0 by writing:
```julia
Nothing = 1.0
@model ...
```
It's not documented anywhere that users of the macro aren't allowed to redefine `Nothing` before calling it. I think the readability issue can be addressed by adding a comment explaining why the types need to be prefixed.

blegat:
Then we might also need to prefix `Dict`.

mlubin:
> Then we might also need to prefix Dict.

Fine with me.

blegat:
The issue is that we are escaping everything:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/00b5965ea1491a893af01c0bf8b5501a0f11c4bb/src/Utilities/model.jl#L648
Otherwise, it would automatically be prefixed by MOIU.

blegat:
Superseeded by https://github.com/JuliaOpt/MathOptInterface.jl/pull/544

Issue #122, opened 2018-11-07
Helpful error instead of segfault for empty model


blegat:
With an empty model, SCS segfaults. Instead it should throw an helpful error.
See https://github.com/JuliaOpt/JuMP.jl/issues/1502

kalmarek:
is a test that `n>0` (e.g. in https://github.com/JuliaOpt/SCS.jl/blob/master/src/c_wrapper.jl#L39) sufficient?

blegat:
Yes, and also check that `m` is greater than zero.

PR #123, opened 2018-11-08
Fix/empty model

kalmarek:
fixes https://github.com/JuliaOpt/SCS.jl/issues/122
https://github.com/JuliaOpt/JuMP.jl/issues/1502

Two tests in `solve_blank_obj` fail (probably because we `throw`)

btw I noticed this:
```
julia> m = Model(); @variable(m, x <=1); m
Feasibility problem with:
 * 0 linear constraints
 * 1 variable
Solver is default solver

julia> m = Model(); x = @variable(m, x); @constraint(m, x <=1); m
Feasibility problem with:
 * 1 linear constraint
 * 1 variable
Solver is default solver
```
but I guess that it's fixed in MOI due to changes in models

### DIFF
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -45,6 +45,9 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
         slack::Vector{Float64}=Float64[];
         options...)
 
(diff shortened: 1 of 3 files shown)


mlubin:
> btw I noticed this:

That's expected. Variable bounds are handled differently than linear constraints.

PR #144, opened 2019-05-25
Updates to MOI v0.9

blegat:
Also drop Julia v0.6, Julia v0.7 and transition from REQUIRE to Project.toml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: julia
+coveralls: true
 os:
   - linux
   - osx
@@ -9,5 +10,3 @@ notifications:
   email: false
 before_script:
   - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
-after_success:
-  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
-  - 0.7
   - 1.0
+  - 1.1
 notifications:
   email: false
+before_script:
+  - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
 after_success:
-- julia -e 'cd(Pkg.dir("SCS")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: julia
-coveralls: true
 os:
   - linux
   - osx
@@ -9,3 +8,5 @@ julia:
   - 1.2
 notifications:
   email: false
+after_success:
+  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ os:
 julia:
   - 1.0
   - 1.1
+  - 1.2
 notifications:
   email: false
-before_script:
-  - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,25 @@
+name = "SCS"
+uuid = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
+repo = "https://github.com/JuliaOpt/SCS.jl.git"
+version = "0.5.1"
+
+[deps]
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[compat]
+BinaryProvider = "≥ 0.3.0"
+MathOptInterface = "0.9"
+MathProgBase = "~0.5.0, ~0.6, ~0.7"
+julia = "1"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+
+[targets]
+test = ["Test", "Compat"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SCS"
 uuid = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 repo = "https://github.com/JuliaOpt/SCS.jl.git"
-version = "0.5.1"
+version = "0.6.0"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
--- a/Project.toml
+++ b/Project.toml
@@ -18,8 +18,8 @@ MathProgBase = "~0.5.0, ~0.6, ~0.7"
 julia = "1"
 
 [extras]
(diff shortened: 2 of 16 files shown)


src/SCS.jl
```diff
@@ -14,8 +12,6 @@ else
     error("SCS not properly installed. Please run Pkg.build(\"SCS\") and restart julia")
 end
 
-using Compat
-
 function __init__()
     vnum = VersionNumber(SCS_version())
     depsdir = realpath(joinpath(dirname(@__FILE__),"..","deps"))
```
kalmarek: Below: we need version `v2.0.*`, the not-yet-released `v2.1.0` of scs has additional fields in `SCSSettings` so it's not compatible

src/types.jl
```diff
@@ -169,14 +188,15 @@ mutable struct Solution
     x::Array{Float64, 1}
     y::Array{Float64, 1}
     s::Array{Float64, 1}
+    info::SCSInfo
     status::Symbol
     ret_val::Int
 
-    function Solution(x::Array{Float64, 1}, y::Array{Float64, 1}, s::Array{Float64, 1}, ret_val::Int)
+    function Solution(x::Array{Float64, 1}, y::Array{Float64, 1}, s::Array{Float64, 1}, info::SCSInfo, ret_val::Int)
```
kalmarek: please update `status_map` (above) according to `https://github.com/cvxgrp/scs/blob/d3d1d4c49b0aede0ec491fe06c99aef5cfe598c7/include/glbopts.h#L18` (or similar in `MOI_wrapper.jl`)

I once opened a pull with this, but this was in pre-MOI times, so we left it for later, maybe now is the time;
maybe it's a good idea to share the definition with MOI?

src/types.jl
```diff
@@ -169,14 +188,15 @@ mutable struct Solution
...
         if haskey(status_map, ret_val)
-            return new(x, y, s, status_map[ret_val], ret_val)
```
kalmarek: its better to use `get(status_map, ret_val, :UnknownError)` here

Project.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[compat]
+BinaryProvider = "≥ 0.3.0"
+MathOptInterface = "0.9"
+MathProgBase = "~0.5.0, ~0.6, ~0.7"
+julia = "1"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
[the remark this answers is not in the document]
blegat: Yes MPB needs it, see https://github.com/JuliaOpt/MathProgBase.jl/pull/234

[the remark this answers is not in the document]
kalmarek: I don't quite follow: if MPB needs it then MPB should have it listed in its `Project.toml`...?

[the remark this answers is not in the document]
blegat: The issue here is that in MPB, solver should includes files in the `test` folder of MPB. So it is like these files were in the `test` folder of the solver itself so the solver requires Compat too

src/types.jl
```diff
@@ -99,14 +99,13 @@ struct SCSSolution
...
 
-
 struct SCSInfo
     iter::Int
     # We need to allocate 32 bytes for a character string, so we allocate 256 bits
     # of integer instead
     # TODO: Find a better way to do this
-    status1::Int128
-    status2::Int128
+    status1::UInt128
```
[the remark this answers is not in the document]
blegat: I prefer using `Cchar`, since `char` is used in `scs.h`

src/types.jl
blegat: The status should be one of the MPB statuses. For custom status described by SCS, this should be `raw_status(info)`.
We should remove this from the struct since it is tied to MPB

blegat: Ah no I see, we don't handle outside [-4, 1]. Let's leave it for a separate since I moved it to MPB_wrapper

src/types.jl
```diff
@@ -119,26 +114,13 @@ struct SCSInfo
...
-            status >>= 8
-        end
-        if eos
-            break
-        end
-    end
-    return String(chars)
+    s = collect(info.status)
+    len = findfirst(iszero, s) - 1
+    # There is no method String(::Vector{Cchar}) so we convert to `UInt8`.
```
kalmarek: you could `String(reinterpret(UInt8, s)[1:len])` as well

Project.toml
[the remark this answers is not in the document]
kalmarek: I see.
I'd appreciate a comment somewhere in tests that `Compat` is needed there (so that it's not left dangling when MPB support is removed)

src/types.jl
blegat: If we use the fact that `UInt8` has the same binary representation of `Cchar`, we might as well store it as `UInt8` in the struct

kalmarek: yes, this is why my intial suggestion was to go with `Cuchar`; but this is a matter of naming/taste, so it can stay as it is now 

(but this convert via array comprehension makes me wince :))

blegat: Yes but I would rather have the code not depend on equivalence of bit representation just to have `raw_status` be slightly faster. What's wrong with array comprehension :-P ? How would you write it?

kalmarek: I'm not feeling strongly about it; that bike-shedding was intended as a joke ;-).
I'd probably go for `reinterpret(UInt8, s[1:len])` if we really do care about this one allocation (we don't:).

kalmarek:
@blegat after suggested changes in `c_wrapper.jl` this looks good to me.

kalmarek:
`scs-2.1.1` has been tagged today (https://github.com/cvxgrp/scs/releases/tag/2.1.1) and I have a branch with gpu solver working from julia. It would be nice to have these in before v0.6 (or we will have 0.7 immediately after) @blegat could you postpone the version bump?

blegat:
Not sure sure what you suggest, do you want to merge your branch before and tag v0.6 ?
The version bump is for dropping Julia v0.6 and v0.7 so if you merge your branch, drop Julia v0.6 and 0.7 and then tag v0.6 then this PR can just be v0.6.1

kalmarek:
I thought You're about to merge this soon;

scs-2.1.0 breaks binary compatibility and we have to modify SCSSettings struct accordingly (https://github.com/kalmarek/SCS.jl/commit/2c3c4c0f14e34858386413b04c638e054386ad1e), so that alone would be a reason to bump to v0.6;

gpu-solver is a more extensive one: https://github.com/kalmarek/SCS.jl/commit/58556100b3b3403edcf6da202ee932cd0abd98c2

but if you don't object, I'd prepare pulls and merge these before

blegat:
> but if you don't object, I'd prepare pulls and merge these before

Yes, this PR might be merged in a few days but you should have time to do it before. Just make sure to drop Julia v0.6 and v0.7.

.travis.yml
```diff
@@ -3,10 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
-  - 0.7
   - 1.0
+  - 1.1
```
kalmarek: we could add `1.2` here as well

.travis.yml
```diff
@@ -3,10 +3,11 @@ os:
...
 notifications:
   email: false
+before_script:
+  - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
```
kalmarek: On CI we should use only released versions, shouldn't we?; If the changes in MOI are needed here, this has to wait until we can use the released version

.travis.yml
```diff
@@ -3,10 +3,11 @@ os:
...
 after_success:
-- julia -e 'cd(Pkg.dir("SCS")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
```
kalmarek: I believe that a simple `- coveralls: true` should do, according to docs:
https://docs.travis-ci.com/user/languages/julia/#coverage

appveyor.yml
```diff
@@ -1,8 +1,7 @@
 environment:
   matrix:
-  - julia_version: 0.6
-  - julia_version: 0.7
   - julia_version: 1.0
+  - julia_version: 1.1
```
kalmarek: add `1.2` as above

.travis.yml
blegat: Yes, it will be removed before we merge

.travis.yml
blegat: It worked on MOI and JuMP but it gives a weird error here:
https://travis-ci.org/JuliaOpt/SCS.jl/jobs/571513843#L936-L978
Maybe it works with codecov but not with coveralls

PR #147, opened 2019-07-10
Enh/gpu solver

kalmarek:
* SCS on gpu uses cuda which requires `scs_int == int`, hence DLONG=0 and we need all structs passed to scslibs parametrized by `Int32` for proper alignment;
* detection for present `libscs*` in custom installation
* a example for compilation of `scslib`s is added to README
* we can't really test this on CI

please share your thoughts if this can be supported


kalmarek:
@blegat this is failing only for julia-0.6 so should be safe to (rebase and) merge after #144

blegat:
@kalmarek Feel free to drop Julia v0.6 on this PR and I'll rebase #144 or the other way around, as you prefer

kalmarek:
@blegat I don't think that cherry-picking the dropping of v0.6 from your branch makes much sense. Lets wait with this one until #144 is merged

kalmarek:
@juan-pablo-vielma I'm not sure, but isn't `long` on windows x86 `4 bytes` long? then DLONG=1 does not enforce Int64...

kalmarek:
before we merge this:

* this introduces `const SCS.available_solvers::Vector{SCS.LinearSolver}`; I'm not supper happy with the name, but it is what it says.
* I changed the names of `SCS.Direct` → `SCS.DirectSolver` (the same for indirect), I think it makes `linear_solver=SCS.DirectSolver` more readable, and these options are easier to find now;
* most of the structs are parametrised by `T<:SCSInt`, since gpu stuff needs to be compiled with `DLONG=0`, even for `x86_64`
* as said at the beginning, there is currently no way of testing the gpu part on Travis;

juan-pablo-vielma:
@kalmarek can you give me more details on the DLONG issue. Might be worth testing on some GPUs before merging. What would be the hardware requirements for this?

kalmarek:
@juan-pablo-vielma I guess that the DLONG issue is probably only because of my wrong assumptions: I thought that regardless of arch we want `scs` ints to be 8 bytes long (hence `DLONG=1`). However, `scs_int` is either `long` (if `DLONG=1`) or `int` (if `DLONG=0`) and for 32-bit windows these two are the same, so I needed to change (in julia) `scsint_t` from `Int64` to `Int`.

To test this you need a CUDA-enabled gpu
1. Clone `scs` (there are some latest fixes on master)
2. Set `CUDA_PATH` (usually done by the cuda installer/distribution)
3. inside `scs` compile scs libs linking against julia's `libopenblas` e.g. [like this](https://gist.github.com/kalmarek/1ddc06f2d76f6b6c343bf7cb9be98d75).
4. in julia set `ENV["JULIA_SCS_LIBRARY_PATH"]="<path/to/scs>/out/"`
5. `using Pkg; pkg"build SCS"; pkg"test SCS"`

bodono:
I've been having a lot of trouble getting the types right for windows since I don't have a windows machine to test on and windows is very confusing with respect to integer types. Currently when `DLONG=1` on windows 64 I set it to `int64_t` [here](https://github.com/cvxgrp/scs/blob/master/include/glbopts.h#L99), for win 32 it appears I just use the standard logic which sounds like it's wrong if I'm reading your message correctly. I'm happy to change it / accept a PR if you know the right types to use here.

kalmarek:
@bodono I honestly have no idea what is right and what is wrong in this context ;-)
I'm not a C/C++ programmer, and never had to deal with portable (multiple platforms) code. Here I just assumed that `long` will always be 8 bytes long integer (`Int64` in julia); this assumption was wrong and lead to crashes on win-32. 

Maybe we should test this on linux-x86 as well? (binaries for scs are build anyway)

kalmarek:
It was easier to rewrite everything than rebasing (or my git-fu sucks:)

---

I've just confirmed that this works with `scs-2.1.1` build from source with gpu support. There I have
```julia
julia> SCS.available_solvers
3-element Array{DataType,1}:
 SCS.DirectSolver     
 SCS.IndirectSolver   
 SCS.IndirectGpuSolver
```
and tests are run (and pass) for all of these solvers.

* on `scs-2.1.2` there will be (probably) `DirectGpuSolver` as well: https://github.com/cvxgrp/scs/commit/82df6213af31c87fccc23bc4e0b71b58a2cb7c9f but that should work with minimal changes to `build.jl` and `available_solvers`.

kalmarek:
@blegat @mlubin can we merge this before `v0.7` is tagged? #177 already bumps the version

mlubin:
@kalmarek You now have commit access, so merge cautiously :)

kalmarek:
this is exactly why I always like to have another pair of eyes look over my changes;)
but if you're ok with the changes i can merge it

kalmarek:
ok, so I just checked that the provided instructions work and tests pass on a cuda cluster nearby, so I'll move forward and merge this.

any chances of cuda on our ci?

ericphanson:
JuliaGPU provides some gpu CI resources: https://github.com/JuliaGPU/gitlab-ci

By the way, [CuArrays now uses artifacts to load cuda](https://github.com/JuliaGPU/CuArrays.jl/commit/85145694cbbb2cdbfae04f8a04cf56704cac529e); I wonder if it's possible to avoid requiring users install CUDA and build SCS themselves.

kalmarek:
maybe something could be done for additional SCSgpu artifact (and `Requires.jl`) but given how huge cuda is (~400MB) to how small scs is (~250KB) (that's not a typo:) I don't think we should do it by default

I'll have a look at the CI, thanks for the info

ericphanson:
> maybe something could be done for additional SCSgpu artifact (and `Requires.jl`) but given how huge cuda is (~400MB) to how small scs is (~250KB) (that's not a typo:) I don't think we should do it by default

Sure, that makes sense. Maybe one thing to look at is how CuArrays itself does it (I think some of the relevant code is in https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/bindeps.jl). They seem to use artifacts directly rather than JLL packages, taking advantage of the fact that artifacts can be marked lazy, so they don't get downloaded until they are needed, and then can be loaded at runtime. Following this path, I think you could have a function `load_gpu_binaries()` that loads the artifacts (downloading them if they haven't been already). That could avoid the need for Requires. I haven't done this myself though, maybe there is a catch I'm missing.

One other good thing about artifacts (though this is true regardless if they are used via JLL packages or not) is that they are content-addressed and stored in the users `~/.julia/artifacts` folder, so if a user installs and uses another package that uses cuda via artifacts, e.g. CuArrays, and thus gets a cuda artifact, and then later they call the hypothetical `load_gpu_binaries()` in SCS, it shouldn't have to redownload anything or use extra disk space.

ericphanson:
I asked about this a bit on the #gpu channel in the Julia slack, and @maleadt was kind enough to provide some help. There's currently a PR https://github.com/JuliaPackaging/Yggdrasil/pull/656 to provide a massive `CUDA_full_jll` that is to be used at build time (not downloaded to the user's computer). We could have a `SCS_CUDA_jll` which is built (in Yggdrasil) using this `CUDA_full_jll` at build time, though it may also need to link against a `libcuda_jll`. I think to get the lazy-loading functionality we may still need to use artifacts manually or requires though.

Here's the full conversation, since Slack has a very short history:

>Eric Hanson Today at 15:42
>I noticed the new cuda-via-artifacts support in CuArrays has a fair bit of code involved in getting the right version of coda and all (https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/bindeps.jl). If other packages want to load cuda via artifacts, possibly lazily, (ref: https://github.com/JuliaOpt/SCS.jl/pull/147#issuecomment-606039317), do they need to duplicate that? Since the cuda artifacts are big, it seems to make sense to try to use the exact same versions in many packages so the user doesn't end up needing to download a lot and use a lot of disk space. Could it make sense to extract out the cuda artifact loading to a separate package so that loading code and versioning would be centralized, preventing a proliferation of downstream versions?
>12 replies
>Tim Besard  19 minutes ago
>I ultimately hope that the BB functionality to generate JLL packages will be flexible enough to be able to create a libcuda_jll
>Tim Besard  19 minutes ago
>in the mean time we could try and replace the libcuda_jll entry points with that logic to make it possible to reuse that code, yesx
>Tim Besard  18 minutes ago
>for now you can use call CuArrays.libcublas(), but that's not a public API of course
>Eric Hanson  10 minutes ago
>Ok, makes sense. I think the package in question, SCS, might not want to depend on CuArrays since most of its use is cpu based (it just got gpu support very recently). But I think also the SCS binary might need to be built against cuda, so I think there would be a SCS_cuda_jll that depends on cuda artifacts. I don’t really know how it all works though.
>Tim Besard  9 minutes ago
>if it needs to build against CUDA, that needs to happen in a BB recipe where you (will be able to) depend in CUDA_full_jll giving you the entire CUDA toolkit (including nvcc, which isn't included in CUDnative/CuArrays' artifacts)
>Tim Besard  9 minutes ago
>https://github.com/JuliaPackaging/Yggdrasil/pull/656
>Tim Besard  9 minutes ago
>that artifact isn't for use by packages though
>Tim Besard  9 minutes ago
>it's huge
>Eric Hanson  7 minutes ago
>Ok, that sounds good! Maybe SCS.jl doesn’t need to depend on cuda at all then but rather just a SCS_cuda_jll that uses that CUDA_full_jll in the recipe
>Tim Besard  6 minutes ago
>do note that the built SCS_cuda_jll might then depend on libcuda, so it might still need libcuda_jll, unless you decide to link statically or redistribute those libraries too
>Tim Besard  6 minutes ago
>but let's think about that when we you have something that builds :slightly_smiling_face:
>Eric Hanson  5 minutes ago
>ah okay, makes sense, thanks for the help!

PR #149, opened 2019-07-13
Scs 2.1.1

kalmarek:


kalmarek:
If you don't mind @juan-pablo-vielma I copied your changes to build.jl, since cherry-picking was not an option

kalmarek:
@juan-pablo-vielma can you have a look at mac failures?

juan-pablo-vielma:
Ok looks like this is because of the change from     "https://github.com/JuliaLinearAlgebra/OpenBLASBuilder/releases/download/v0.3.0-3/build_OpenBLAS.v0.3.0.jl"
to 
 "https://github.com/JuliaPackaging/Yggdrasil/releases/download/OpenBLAS-v0.3.5-2/build_OpenBLAS.v0.3.5.jl"

for the dependencies in SCSBuilder (the error also happens for v2.0.2 with the updated dependencies).

What was the reason for that change?

kalmarek:
I changed this since its the official OpenBlasBuilder and https://github.com/JuliaLang/julia/pull/31441 has been merged for `julia-1.2` so i believe it's way to go

kalmarek:
@juan-pablo-vielma could you have a look at mac failures again? for some reason 1.2 works, but 1.0 and 1.1 can't satisfy `:indirect` `LibraryProduct`...

juan-pablo-vielma:
@kalmarek a small change to the `libopenblas64_` rename fixed the issues for OSX. 
I created a PR to kalmarek:scs-2.1.1 with the updates. 
Unfortunately, Windows is still failing and I do not have good leads on that.

kalmarek:
ok, now everything fails ;-) but due to MOI changes;
(I guess MOI version should be capped, since it's moving so fast)

Surprisingly, windows x86 hits different error than linux/macos...

I will ask some of the local magicians to have a look at windows x64 in the next week

blegat:
> ok, now everything fails ;-) but due to MOI changes;
> (I guess MOI version should be capped, since it's moving so fast)

Should be resolved by https://github.com/JuliaOpt/SCS.jl/pull/159. SCS minor version is capped but you cannot cap the patch version. It's only new tests being added, SCS was not broken.

kalmarek:
@blegat I was thinking about capping MOI version in SCS's Project.toml, but you're right that minor version bumps in MOI could brake tests here, so there's nothing we could/should do about it here

juan-pablo-vielma:
I created another PR to kalmarek:scs-2.1.1 that fixes windows x64
Not sure what will happen with windows x86

kalmarek:
Many thanks! Lets see what happens for windows now...

kalmarek:
@blegat could we merge this and wait for #147 to release as 0.7 ? the other pull is also breaking.

blegat:
> could we merge this and wait for #147 to release as 0.7 ?

Sounds good, I created a milestone and added #147 in it to make sure we wait for it for v0.7.

PR #152, opened 2019-08-16
Allow CI failure with Julia 1.2

rschwarz:
We can still see the outcome, but GitHub does not consider it a failed test.
I think this makes sense, since Julia 1.2 is not released yet?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ julia:
   - 1.0
   - 1.1
   - 1.2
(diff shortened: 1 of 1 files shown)


rschwarz:
I wanted to "suggest" this to #144 but didn't know how, so I created a separate PR.

blegat:
The failure seems to be due to https://github.com/JuliaLang/julia/issues/32167

PR #158, opened 2019-10-08
Fix solve_time computation

mlubin:
The fields in the SCS struct are in milliseconds. The SolveTime
attribute is in seconds.
https://github.com/cvxgrp/scs/blob/e5bb794ac014b7a86d127ac03651d2c8a12ecba8/include/scs.h#L98


mlubin:
Tests pass locally. Failures are due to unrelated breakage from MOI 0.9.4.

PR #160, opened 2019-10-11
expose the number of iterations as an MOI attribute

mlubin:
About naming the attribute: SCS does multiple kinds of iterations, e.g., ADMM iterations (main outer iterations) and CG iterations (inner iterations, in indirect mode only), so being specific seems reasonable. It's not obvious that this could or should be covered by a new MOI attribute, so I made it specific to SCS.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -5,6 +5,16 @@ const VI = MOI.VariableIndex
 
 const MOIU = MOI.Utilities
 
+"""
+    ADMMIterations()
+
+The number of ADMM iterations completed during the solve.
+"""
+struct ADMMIterations <: MOI.AbstractModelAttribute
+end
+
+MOI.is_set_by_optimize(::ADMMIterations) = true
+
 mutable struct MOISolution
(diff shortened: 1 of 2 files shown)


src/MOI_wrapper.jl
```diff
@@ -5,6 +5,16 @@ const VI = MOI.VariableIndex
 
 const MOIU = MOI.Utilities
 
+"""
+    ADMMIterations()
```
blegat: Maybe `NumberOfADMMIterations` to be consistent with other MOI attributes ?

mlubin: As a counter point,
http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.SimplexIterations
http://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.BarrierIterations

(Maybe we should remove or refactor those attributes. They're very specific to MIP solvers.)

blegat: Ok, then `ADMMIterations` is better indeed :)

Issue #163, opened 2019-10-25
Segmentation fault with custom-built SCS


riccardomurri:
Trying to run the same test code as in #162 but with a custom-built
version of SCS results in an instant segmentation fault:

```
┌ Info: Creating convex optimization problem ...
│   size(A) = (1016, 4064)
│   size(x) = (4064, 1)
│   size(b) = (1016,)
│   size(C) = (4064, 4064)
└   size(D) = (4064, 4064)
┌ Debug: Done at 2019-10-25T20:28:01.338
└ @ Main ~/test.jl:38
[ Info: Solving optimization problem ...

signal (11): Segmentation fault
in expression starting at /home/ubuntu/test.jl:12
jl_gc_pool_alloc at /buildworker/worker/package_linux64/build/src/gc.c:1120
jl_gc_alloc_ at /buildworker/worker/package_linux64/build/src/julia_internal.h:274 [inlined]
jl_gc_alloc at /buildworker/worker/package_linux64/build/src/gc.c:2949
_new_array_ at /buildworker/worker/package_linux64/build/src/array.c:100 [inlined]
_new_array at /buildworker/worker/package_linux64/build/src/array.c:160 [inlined]
jl_alloc_array_1d at /buildworker/worker/package_linux64/build/src/array.c:420
ml_matches_visitor at /buildworker/worker/package_linux64/build/src/gf.c:2502
jl_typemap_intersection_node_visitor at /buildworker/worker/package_linux64/build/src/typemap.c:472
jl_typemap_intersection_visitor at /buildworker/worker/package_linux64/build/src/typemap.c:567
ml_matches at /buildworker/worker/package_linux64/build/src/gf.c:2547
cache_method at /buildworker/worker/package_linux64/build/src/gf.c:946
cache_method at /buildworker/worker/package_linux64/build/src/gf.c:1045 [inlined]
jl_mt_assoc_by_type at /buildworker/worker/package_linux64/build/src/gf.c:1045
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2153
_ntuple at ./ntuple.jl:36
ntuple at ./ntuple.jl:18 [inlined]
Type at /home/ubuntu/.julia/packages/SCS/Y9wx3/src/types.jl:117
#SCS_solve#5 at /home/ubuntu/.julia/packages/SCS/Y9wx3/src/c_wrapper.jl:46
#SCS_solve at ./none:0
[...]

PR #164, opened 2019-11-06
Fix/scs solve

kalmarek:
as noted in https://github.com/JuliaOpt/SCS.jl/issues/163#issuecomment-546693066, the argument `s` was being overwritten by `slack`; this fixes this problem.

I also store a `scsmat::Ref{SCSMatrix}` inside `mmatrix::ManagedSCSMatrix`, so that as long as `mmatrix` lives, the corresponding `scsmat` points to a valid memory. However, since I take this as an opportunity to learn about julia/C interaction, I'd ask someone more experienced with the matter (@juan-pablo-vielma, @mlubin) for review.

Also: `ManagedSCSMatrix` claims to accept `A::AbstractVector`, but of course it's definition (with access to `rowval` and `colptr`) will fail. Should we change the definition of `SCSVecOrMatOrSparse`, or add a new method? (btw. is there a reason why SCS types are exported)?

### DIFF
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -50,15 +50,17 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
     end
     solution = SCSSolution(pointer(primal), pointer(dual), pointer(slack))
 
-    managed_matrix = ManagedSCSMatrix(m, n, A)
-    matrix = Ref(SCSMatrix(managed_matrix))
-    settings = Ref(SCSSettings(T; options...))
-    data = Ref(SCSData(m, n, Base.unsafe_convert(Ptr{SCSMatrix}, matrix), pointer(b), pointer(c), Base.unsafe_convert(Ptr{SCSSettings},settings)))
+    settings = Base.cconvert(Ref{SCSSettings}, SCSSettings(T; options...))
+    mmatrix = ManagedSCSMatrix(m, n, A)
+    data = Base.cconvert(Ref{SCSData}, SCSData(m, n,
+        Base.unsafe_convert(Ref{SCSMatrix}, mmatrix.scsmat), #creates Ptr{SCSMatrix}
+        pointer(b), pointer(c),
+        Base.unsafe_convert(Ref{SCSSettings}, settings))) #creates Ptr{SCSSettings}
+    # unsafe_convert doesn't protect from GC: mmatrix and settings must be GC.@preserved
+    cone = Base.cconvert(Ref{SCSCone}, SCSCone(f, l, q, s, ep, ed, p))
+    info = Base.cconvert(Ref{SCSInfo}, SCSInfo())
 
-    cone = Ref(SCSCone(f, l, q, s, ep, ed, p))
-    info = Ref(SCSInfo())
-
-    Base.GC.@preserve managed_matrix matrix settings b c q s p begin
+    Base.GC.@preserve mmatrix settings b c q s p begin
         p_work = SCS_init(T, data, cone, info)
         status = SCS_solve(T, p_work, data, cone, solution, info)
         SCS_finish(T, p_work)
@@ -82,7 +84,7 @@ for (T, lib) in zip([SCS.Direct, SCS.Indirect], [SCS.direct, SCS.indirect])
         function SCS_init(::Type{$T}, data::Ref{SCSData}, cone::Ref{SCSCone}, info::Ref{SCSInfo})
 
             p_work = ccall((:scs_init, $lib), Ptr{Nothing},
-                (Ptr{SCSData}, Ptr{SCSCone}, Ptr{SCSInfo}),
+                (Ref{SCSData}, Ref{SCSCone}, Ref{SCSInfo}),
                 data, cone, info)
 
             return p_work
@@ -92,7 +94,7 @@ for (T, lib) in zip([SCS.Direct, SCS.Indirect], [SCS.direct, SCS.indirect])
         function SCS_solve(::Type{$T}, p_work::Ptr{Nothing}, data::Ref{SCSData}, cone::Ref{SCSCone}, solution::SCSSolution, info::Ref{SCSInfo})
 
             status = ccall((:scs_solve, $lib), Int,
-                (Ptr{Nothing}, Ptr{SCSData}, Ptr{SCSCone}, Ref{SCSSolution}, Ptr{SCSInfo}),
+                (Ptr{Nothing}, Ref{SCSData}, Ref{SCSCone}, Ref{SCSSolution}, Ref{SCSInfo}),
                 p_work, data, cone, solution, info)
 
             return status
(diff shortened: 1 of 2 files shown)


kalmarek:
based on discussion: https://discourse.julialang.org/t/how-to-keep-a-reference-for-c-structure-to-avoid-gc/9310/25 I rewrote the handling of/conversion to pointers

kalmarek:
@mlubin, @blegat could you have a look at this and provide feedback?

src/c_wrapper.jl
```diff
@@ -37,35 +37,36 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
     ws = (:warm_start=>true) in options
 
-    if ws && length(primal_sol) == n && length(dual_sol) == m && length(slack) == m
-        x = primal_sol
-        y = dual_sol
-        s = slack
+    if ws && length(primal_sol) == n && length(dual_sol) == length(slack) == m
+        primal = primal_sol
+        dual = dual_sol
+        slack = slack
```
blegat: This line can be removed, it has no effect

kalmarek: which one?
Do you mean that the default `zeros(...)` for `primal`, `dual` and `slack` should be the default values?

kalmarek: how about something like this:
```julia
function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
        m::Int, n::Int, A::SCSVecOrMatOrSparse, b::Array{Float64},
        c::Array{Float64}, f::Int, l::Int, q::Array{Int}, s::Array{Int},
        ep::Int, ed::Int, p::Array{Float64},
        primal_sol::Vector{Float64}=zeros(n),
        dual_sol::Vector{Float64}=zeros(m),
        slack::Vector{Float64}=zeros(m);
        options...)

    n > 0 || throw(ArgumentError("The number of variables in SCSModel must be greater than 0"))
    m > 0 || throw(ArgumentError("The number of constraints in SCSModel must be greater than 0"))

    ws = (:warm_start=>true) in options
    
    if !(length(primal_sol) == n && length(dual_sol) == length(slack) == m)
        ws && @warn "The provided warmstart doesn't match problem sizes; discarding!"
        primal_sol = zeros(n)
        dual_sol = zeros(m)
        slack = zeros(m)
    else
        if ws # the provided warmstart matches problem sizes and user asked for warmstaring
            nothing
        else 
            primal_sol = fill!(primal_sol, 0.0)
            dual_sol = fill!(dual_sol, 0.0)
            slack = fill!(slack, 0.0)
        end
    end
    
    solution = SCSSolution(pointer(primal_sol), pointer(dual_sol), pointer(slack))
    (...)
    return Solution(primal_sol, dual_sol, slack, info[], status)
end
```

blegat: I simply mean that `slack = slack` is a no-op

mlubin:
Why is Julia 1.2 an allowed failure on travis? Should we also test 1.3?

src/types.jl
```diff
@@ -17,26 +17,20 @@ struct ManagedSCSMatrix
     values::Vector{Cdouble}
     rowval::Vector{Int}
     colptr::Vector{Int}
-    m::Int
-    n::Int
-end
+    scsmat::Base.RefValue{SCSMatrix}
```
mlubin: RefValue isn't mentioned anywhere in the documentation (https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/), are we supposed to be using this type?

src/c_wrapper.jl
```diff
@@ -29,43 +29,49 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
+        if !ws
+            primal_sol = fill!(primal_sol, 0.0)
+            dual_sol = fill!(dual_sol, 0.0)
+            slack = fill!(slack, 0.0)
+        end
     else
-        x = zeros(n)
-        y = zeros(m)
-        s = zeros(m)
+        ws && @warn "The provided warmstart doesn't match problem sizes; discarding!"
```
mlubin: IMO it would be fine to turn this into an error case. Providing vectors of the wrong length probably means something bad happened.

src/c_wrapper.jl
```diff
@@ -29,43 +29,49 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
     ws = (:warm_start=>true) in options
+    sizes_correct = length(primal_sol) == n && length(dual_sol) == length(slack) == m
 
-    if ws && length(primal_sol) == n && length(dual_sol) == m && length(slack) == m
-        x = primal_sol
-        y = dual_sol
-        s = slack
+    if sizes_correct
+        if !ws
+            primal_sol = fill!(primal_sol, 0.0)
```
mlubin: `fill!(primal_sol, 0.0)` (assignment isn't needed). Same for the other two.

src/c_wrapper.jl
```diff
@@ -29,43 +29,49 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
+        dual_sol = zeros(m)
+        slack = zeros(m)
     end
-    solution = SCSSolution(pointer(x), pointer(y), pointer(s))
 
-    Base.GC.@preserve managed_matrix matrix settings b c q s p begin
+    solution = SCSSolution(pointer(primal_sol), pointer(dual_sol), pointer(slack))
+
+    settings = Base.cconvert(Ref{SCSSettings}, SCSSettings(T; options...))
+    mmatrix = ManagedSCSMatrix(m, n, A)
```
mlubin: Naming: `managed_matrix`. Use only standard abbreviations (http://www.juliaopt.org/JuMP.jl/dev/style/#Abbreviations-1).

src/types.jl
kalmarek: it's the return value of `Base.cconvert(Ref{A}, A)`;
Note that we need to ccall with a `Ref` to `SCSData` which contains a `Ptr{SCSMatrix}`. All of the examples given in the docs pass c-compatible julia structs.

However, its just a way of encapsulating the object and its C-compatible reference for `GC.@protect`ed calls. Maybe the complication isn't worth it.

src/c_wrapper.jl
kalmarek: should it be is a `DomainError` or `ArgumentError`?

blegat: I'd say `ArgumentError`

src/types.jl
mlubin: It's ok to leave it in, but for the benefit of others reading the code, could you add comments explaining why we need to use an undocumented type here?

src/c_wrapper.jl
kalmarek: that's true, but only since these are `Vector`s (we require this in the signature) and these are mutable. Is it guaranteed by any means?
I thought the best practice was to use only the return value of the mutating call. But sure, I can change this if You wish.

kalmarek:
@mlubin I will split this pull into two and then we can discuss the refactoring only.

PR #168, opened 2019-12-22
Fix/scs solve new

kalmarek:
fix accidental override of an argument;
simplify warmstart logic in SCS_solve

### DIFF
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -42,12 +42,12 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
 
     if warmstart_sizes_are_correct
         if !ws
-            primal_sol = fill!(primal_sol, 0.0)
-            dual_sol = fill!(dual_sol, 0.0)
-            slack = fill!(slack, 0.0)
+            fill!(primal_sol, 0.0)
+            fill!(dual_sol, 0.0)
+            fill!(slack, 0.0)
         end
     else
-        ws && throw(ArgumentError("The provided warmstart doesn't match problem sizes; discarding!"))
+        ws && throw(ArgumentError("The provided warmstart doesn't match problem sizes"))
         primal_sol = zeros(n)
         dual_sol = zeros(m)
         slack = zeros(m)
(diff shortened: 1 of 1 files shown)


src/c_wrapper.jl
```diff
@@ -29,14 +29,32 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
+    warmstart_sizes_are_correct = length(primal_sol) == n && length(dual_sol) == length(slack) == m
+
+    if warmstart_sizes_are_correct
+        if !ws
+            primal_sol = fill!(primal_sol, 0.0)
+            dual_sol = fill!(dual_sol, 0.0)
+            slack = fill!(slack, 0.0)
+        end
+    else
+        ws && throw(ArgumentError("The provided warmstart doesn't match problem sizes; discarding!"))
```
mlubin: "discarding" is inaccurate. The execution will stop here.

src/c_wrapper.jl
```diff
@@ -29,14 +29,32 @@ function SCS_solve(T::Union{Type{Direct}, Type{Indirect}},
...
 
     n > 0 || throw(ArgumentError("The number of variables in SCSModel must be greater than 0"))
     m > 0 || throw(ArgumentError("The number of constraints in SCSModel must be greater than 0"))
 
+    ws = (:warm_start=>true) in options
+    warmstart_sizes_are_correct = length(primal_sol) == n && length(dual_sol) == length(slack) == m
+
+    if warmstart_sizes_are_correct
+        if !ws
+            primal_sol = fill!(primal_sol, 0.0)
```
mlubin: Continuing the discussion from the last PR (https://github.com/JuliaOpt/SCS.jl/pull/164/files#r360657916):

> that's true, but only since these are Vectors (we require this in the signature) and these are mutable. Is it guaranteed by any means?

`fill!` is guaranteed to mutate its first argument by definition.

> I thought the best practice was to use only the return value of the mutating call. But sure, I can change this if You wish.

Where is this practice discussed? I haven't seen it before, and it doesn't make much sense to me. What if a function mutates multiple arguments?

src/c_wrapper.jl
kalmarek: thanks for catching this, fixed

src/c_wrapper.jl
kalmarek: This is just a side note:
I thought that this is the standard practice, e.g. with mutable arithmetic types when writing generic, fast (allocationless) code which works for both mutable and immutable types: e.g. you need to write `z = mul!(z, x, y)` since just `mul!(z, x, y)` while correct for mutables, will discard the value when `z` can not be mutated, i.e. mutability is an implementation detail.

Mutability of `Vectors` will not change in foreseeable future, I completely agree ;)

PR #169, opened 2019-12-23
Enh/refactor scs matrix usage

kalmarek:
according to the discussion https://discourse.julialang.org/t/how-to-keep-a-reference-for-c-structure-to-avoid-gc/9310/25 the way from `M::SCSMatrix` to `Ptr{SCSMatrix}` should be either
```julia
M_ref = Base.cconvert(Ref{SCSMatrix}, M) # Base.RefValue{SCSMatrix}
M_ptr = Base.unsafe_convert(Ref{SCSMatrix}, M_ref) # Ptr{SCSMatrix}, ready to go for ccall
```
or
```julia
M_ref = Base.cconvert(Ptr{SCSMatrix}, Ref(M)) # Base.RefValue{SCSMatrix}
M_ptr = Base.unsafe_convert(Ptr{SCSMatrix}, M_ref) # Ptr{SCSMatrix}
```
This is to align all of the calls / conversions in the code and do not use direct calls to `Ref` outside of `ccall`s.

Additionally (as explained in the comment in the code), `managed_matrix::ManagedSCSMatrix` stores `scsmatref`, a valid reference to its `SCSMatrix` counterpart, so that as long as `managed_matrix` lives, all pointers in `scsmatref` are guaranteed to be valid.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -17,25 +17,19 @@ struct ManagedSCSMatrix
     values::Vector{Cdouble}
     rowval::Vector{Int}
     colptr::Vector{Int}
-    m::Int
-    n::Int
+    scsmat::Base.RefValue{SCSMatrix}
+
+    function ManagedSCSMatrix(m::Integer, n::Integer, A::SCSVecOrMatOrSparse)
+        A_sparse = sparse(A)
+
+        values = copy(A_sparse.nzval)
+        rowval = convert(Array{Int, 1}, A_sparse.rowval .- 1)
+        colptr = convert(Array{Int, 1}, A_sparse.colptr .- 1)
+
+        scsmat = SCSMatrix(pointer(values), pointer(rowval), pointer(colptr), m, n)
+        return new(values, rowval, colptr, Base.cconvert(Ref{SCSMatrix}, scsmat))
+    end
 end
-
-function ManagedSCSMatrix(m::Int, n::Int, A::SCSVecOrMatOrSparse)
-    A_sparse = sparse(A)
-
-    values = copy(A_sparse.nzval)
-    rowval = convert(Array{Int, 1}, A_sparse.rowval .- 1)
-    colptr = convert(Array{Int, 1}, A_sparse.colptr .- 1)
-
-    return ManagedSCSMatrix(values, rowval, colptr, m, n)
-end
-
-# Returns an SCSMatrix. The vectors are *not* GC tracked in the struct.
-# Use this only when you know that the managed matrix will outlive the SCSMatrix.
-SCSMatrix(m::ManagedSCSMatrix) =
-    SCSMatrix(pointer(m.values), pointer(m.rowval), pointer(m.colptr), m.m, m.n)
-
 
 struct SCSSettings
     normalize::Int # boolean, heuristic data rescaling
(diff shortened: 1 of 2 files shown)


kalmarek:
could we work to merge this, then I can update #147 (which is breaking) and release `v0.7`

src/types.jl
```diff
@@ -17,25 +17,33 @@ struct ManagedSCSMatrix
...
+        colptr = convert(Vector{Int}, A_sparse.colptr .- 1)
+
+        # scsmatref holds the reference to SCSMatrix created out of data in ManagedSCSMatrix.
+        # this way the reference to SCSMatrix always points to valid data
+        # as long as the ManagedSCSMatrix is not GC collected.
+        # One MUST
+        #   `Base.unsafe_convert(Ref{SCSMatrix}, scsmatref)`
+        # when assignig to a field of type `Ptr{SCSMatrix}`, or specify
+        #   `Ref{SCSMatrix}`
+        # in in the type tuple when ccalling with `scsmatref`
```
mlubin: Nit: in in

src/types.jl
```diff
@@ -17,25 +17,33 @@ struct ManagedSCSMatrix
...
+        values = convert(Vector{Cdouble}, copy(A_sparse.nzval))
+        rowval = convert(Vector{Int}, A_sparse.rowval .- 1)
+        colptr = convert(Vector{Int}, A_sparse.colptr .- 1)
+
+        # scsmatref holds the reference to SCSMatrix created out of data in ManagedSCSMatrix.
+        # this way the reference to SCSMatrix always points to valid data
+        # as long as the ManagedSCSMatrix is not GC collected.
+        # One MUST
+        #   `Base.unsafe_convert(Ref{SCSMatrix}, scsmatref)`
+        # when assignig to a field of type `Ptr{SCSMatrix}`, or specify
```
mlubin: Typo in assigning

src/types.jl
```diff
@@ -17,25 +17,33 @@ struct ManagedSCSMatrix
...
-end
+        values = convert(Vector{Cdouble}, copy(A_sparse.nzval))
+        rowval = convert(Vector{Int}, A_sparse.rowval .- 1)
+        colptr = convert(Vector{Int}, A_sparse.colptr .- 1)
+
+        # scsmatref holds the reference to SCSMatrix created out of data in ManagedSCSMatrix.
+        # this way the reference to SCSMatrix always points to valid data
+        # as long as the ManagedSCSMatrix is not GC collected.
+        # One MUST
+        #   `Base.unsafe_convert(Ref{SCSMatrix}, scsmatref)`
```
mlubin: What's the reason for presenting two ways to do this instead of recommending the simpler way (specifying Ref in the call type)?

kalmarek: one can use the ref in two different ways 
 * in `ccall` specifying the type of `Ref{SCSMatrix}` (it gets converted to c-compatible `Ptr{SCSMatrx}`), or 
 * by assign it to a `Ptr{SCSMatrix}` field in `SCSData`.

At the moment we do the latter, since `scs_init`, `scs_solve` and `scs_finish` and `ccalls` inside take `Ref{SCSData}`.

kalmarek:
@mlubin ok, let me battle-test this on my "real" examples (large, taking several minutes, enough time to kick GC).

If @blegat doesn't have any objections I'd be happy to merge this.

PR #171, opened 2020-01-21
Scale off-diagonals of b before transforming U -> L

migarstka:
As correctly noted in the `#FIXME comment` the scaling method `scale_coeff` assumes an upper triangular input (triangular numbers yield the diagonal entries of upper triangular elements of a matrix). Therefore, we have to keep the right order with scaling and transforming:

1. Scale, then transform U->L
2. Transform L->U, then unscale

I probably introduced that bug myself in #165. I noticed this when comparing the output of `JuMP + SCS` vs. `SCS.Direct`. I am not sure, why the unit tests didn't catch it.


blegat:
> I am not sure, why the unit tests didn't catch it.

We may be missing a test with PSD matrix of size >= 3 and nonzero constant in off-diagonal entries

PR #177, opened 2020-03-29
Use jll for Julia 1.3 and higher

odow:
x-ref: https://github.com/JuliaPackaging/Yggdrasil/issues/509#issuecomment-605516668

### DIFF
--- a/src/SCS.jl
+++ b/src/SCS.jl
@@ -1,6 +1,6 @@
 module SCS
 
-if VERSION < v"1.3"
+if haskey(ENV, "JULIA_SCS_LIBRARY_PATH") || VERSION < v"1.3"
     if isfile(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
         include(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
     else
--- a/src/SCS.jl
+++ b/src/SCS.jl
@@ -1,17 +1,23 @@
 module SCS
 
-if isfile(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
-    include(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
-else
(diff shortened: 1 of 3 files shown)


kalmarek:
does `jll` artifact allow you to switch the underlying libraries to your own?

odow:
> does jll artifact allow you to switch the underlying libraries to your own?

Not that I'm aware of. I guess we need an environment variable switch somewhere. Thoughts? I guess you're the one who knows the most about SCS.

kalmarek:
yes, but I know next to nothing about artifacts ;-)
I see no changes to `build.jl`, so I believe it will ignore build file and check
https://github.com/JuliaOpt/SCS.jl/blob/bfc756ea4498c89c263f14cdb02dcadd063d17fa/deps/build.jl#L63
therein. Is there any way to bring back the option?

src/SCS.jl
```diff
@@ -1,17 +1,23 @@
 module SCS
 
-if isfile(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
-    include(joinpath(dirname(@__FILE__), "..", "deps", "deps.jl"))
-else
-    error("SCS not properly installed. Please run Pkg.build(\"SCS\") and restart julia")
-end
+if haskey(ENV, "JULIA_SCS_LIBRARY_PATH") || VERSION < v"1.3"
```
odow: @kalmarek will this work?

kalmarek:
so even when `SCS.jl` depends on the artifact (i.e. `julia ^1.3`), when I call `Pkg.build("SCS")` the `deps/build.jl` will be run?

odow:
I think so. Could you try it out?

kalmarek:
works, thanks!

odow:
@kalmarek I just rebased on top of your GPU commits. Assuming CI passes, I'll leave it with you to merge :)

kalmarek:
cool, thanks!

PR #179, opened 2020-05-13
Enh/simplify c wrapper

kalmarek:
@mlubin since you wrote the original code in `c_wrapper.jl` I would like to have a comment from you.

* as noted in commit message `data` and `cone` are marked as `const` in all calls we make to scs:
https://github.com/cvxgrp/scs/blob/0fd7ea85e8b0d878cacf5b1dbce557b330422ff7/include/scs.h#L114 so it's advisable to leave the conversions to `ccall`.
* SCSSolution already contains `Ptr{Cdouble}` (btw: why it was marked as `Nothing=Cvoid`?) so it can be passed directly (but its content needs to outlive it)
* a constructor for creating `SCSData` was added to hide complexity. instead of
```julia
   data = Base.cconvert(Ref{SCSData{T}}, SCSData{T}(m, n,
        Base.unsafe_convert(Ref{SCSMatrix{T}}, managed_matrix.scsmatref), # creates Ptr{SCSMatrix}
        pointer(b), pointer(c),
        Base.unsafe_convert(Ref{SCSSettings{T}}, settings) # creates Ptr{SCSSettings}
        ))
    # unsafe_convert doesn't protect from GC: managed_matrix and settings must be GC.@preserved
end
```
we have now
```julia
data = SCSData(m, n, managed_matrix, b, c, settings)
    # data holds pointers to objects which need to be protected from GC:
    # managed_matrix, b, c and settings
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,14 +58,21 @@ A legacy [MathProgBase](https://github.com/JuliaOpt/MathProgBase.jl) interface i
 ### Options
 All SCS solver options can be set through the direct interface(documented below), through `Convex.jl` or `MathOptInterface.jl`.
 The list of options follows the [`glbopts.h` header](https://github.com/cvxgrp/scs/blob/0fd7ea85e8b0d878cacf5b1dbce557b330422ff7/include/glbopts.h#L30) in lowercase.
-To use these settings you can either pass them as keyword arguments to `SCS_solve` (high level interface) or as arguments to the `SCS.Optimizer` constructor (MathOptInterface), e.g.
+To use these settings you can either pass them as keyword arguments to `SCS_solve` (high level interface) or using the `SCS.Optimizer` constructor (MathOptInterface), e.g.
 ```julia
 # Direct
 solution = SCS_solve(m, n, A, ..., psize; max_iters=10, verbose=0);
 # via MathOptInterface:
-solver = SCS.Optimizer(max_iters=10, verbose=0)
-optimize!(problem, solver) # JuMP
-solve!(problem, solver) # Convex
+optimizer = SCS.Optimizer()
+MOI.set(optimizer, MOI.RawParameter("max_iters"), 10)
+MOI.set(optimizer, MOI.RawParameter("verbose"), 0)
+```
+or via specific helper functions:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,40 +50,47 @@ To switch back to the default binaries delete `JULIA_SCS_LIBRARY_PATH` and call
...
+All SCS solver options can be set through the direct interface(documented below), through `Convex.jl` or `MathOptInterface.jl`.
+The list of options follows the [`glbopts.h` header](https://github.com/cvxgrp/scs/blob/0fd7ea85e8b0d878cacf5b1dbce557b330422ff7/include/glbopts.h#L30) in lowercase.
+To use these settings you can either pass them as keyword arguments to `SCS_solve` (high level interface) or as arguments to the `SCS.Optimizer` constructor (MathOptInterface), e.g.
 ```julia
 # Direct
 solution = SCS_solve(m, n, A, ..., psize; max_iters=10, verbose=0);
-# MathProgBase (with Convex)
-m = solve!(problem, SCSSolver(max_iters=10, verbose=0))
+# via MathOptInterface:
+solver = SCS.Optimizer(max_iters=10, verbose=0)
```
blegat, line 66: This syntax is deprecated. We want to drop keyword argument to the constructor in the future.
It should rather be
```julia
optimizer = SCS.Optimizer()
MOI.set(optimizer, MOI.RawParameter("max_iters"), 10)
MOI.set(optimizer, MOI.RawParameter("verbose"), 0)
```

README.md
```diff
@@ -50,40 +50,47 @@ To switch back to the default binaries delete `JULIA_SCS_LIBRARY_PATH` and call
...
+optimize!(problem, solver) # JuMP
```
blegat, line 67: This is also deprecated. do
```julia
optimizer_constructor = optimizer_with_attributes(SCS.Optimizer, "max_iters" => 10, "verbose" => 0)
set_optimizer(problem, optimizer_constructor)
optimize!(problem)
```

kalmarek:
Thanks for the reviews!

PR #185, opened 2020-07-06
Allow creation of ManagedMatrix without copy of vectors

blegat:
If we are able to directly generate these three vectors from the MOI wrapper and directly call the constructor, it would spare these copies.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -60,7 +60,7 @@ function ManagedSCSMatrix{T}(m::Integer, n::Integer, A::SparseMatrixCSC) where T
     rowval = convert(Vector{T}, A.rowval .- 1)
     colptr = convert(Vector{T}, A.colptr .- 1)
 
-    return ManagedSCSMatrix(m, n, values, rowval, colptr)
+    return ManagedSCSMatrix{T}(m, n, values, rowval, colptr)
 end
 
 function ManagedSCSMatrix{T}(m::Integer, n::Integer, A::AbstractMatrix) where T
--- a/src/types.jl
+++ b/src/types.jl
@@ -33,17 +33,8 @@ struct ManagedSCSMatrix{T<:SCSInt}
     colptr::Vector{T}
     scsmatref::Base.RefValue{SCSMatrix{T}}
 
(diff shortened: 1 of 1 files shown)


src/types.jl
```diff
@@ -60,6 +51,22 @@ struct ManagedSCSMatrix{T<:SCSInt}
...
+    values = convert(Vector{Cdouble}, copy(A.nzval))
+    rowval = convert(Vector{T}, A.rowval .- 1)
+    colptr = convert(Vector{T}, A.colptr .- 1)
+
+    return ManagedSCSMatrix(m, n, values, rowval, colptr)
+end
+
+function ManagedSCSMatrix{T}(m::Integer, n::Integer, A::Matrix) where T
+    return ManagedSCSMatrix(m, n, sparse(A))
+end
```
kalmarek, line 68:
```suggestion
function ManagedSCSMatrix{T}(m::Integer, n::Integer, A::AbstractMatrix) where T
    return ManagedSCSMatrix{T}(m, n, sparse(A))
end
```
`Abstract` is not important, missing `T` is

src/types.jl
```diff
@@ -60,6 +51,22 @@ struct ManagedSCSMatrix{T<:SCSInt}
...
+function ManagedSCSMatrix{T}(m::Integer, n::Integer, A::SparseMatrixCSC) where T
+    # we convert everything to make sure that no conversion will
+    # accidentally happen when ccalling `new`, so that `scsmatref`
+    # holds pointers to actual data stored in the fields.
+
+    values = convert(Vector{Cdouble}, copy(A.nzval))
+    rowval = convert(Vector{T}, A.rowval .- 1)
+    colptr = convert(Vector{T}, A.colptr .- 1)
+
+    return ManagedSCSMatrix(m, n, values, rowval, colptr)
```
kalmarek, line 63:
```suggestion
    return ManagedSCSMatrix{T}(m, n, values, rowval, colptr)
```

kalmarek:
btw. during the move of SCS to jump-dev I think I've lost access rights, i.e. I can't apply the suggestion

mlubin:
@kalmarek Restored your access. It wasn't intentional.

kalmarek:
@mlubin, no problem, thanks!

PR #195, opened 2020-08-11
Explain why installation failed for users with incompatible Julia binary

votroto:
Looking properly at issue https://github.com/jump-dev/SCS.jl/issues/194#issue-673098930, I have realized it is not a problem with `SCS.jl`. The reason for failure to install is caused simply by running unsupported Julia binaries -- such as those linked against OpenBLAS different from the one used by `scsindir.so` from BinaryProvider.

While this is **clearly** mentioned in `README.md`, when a user tries to install `SCS.jl` the easy way first, they get: 
```
ERROR: LoadError: LibraryProduct(nothing, ["libscsindir"], :indirect, "Prefix(/home/user/.julia/packages/SCS/28Svy/deps/usr)") is not satisfied, cannot generate deps.jl!
```
At this point, it is not obvious that it is the users fault.

The user could try to debug the problem by adding `--verbose` to `ARGS`. They would, unfortunately, not be aware that such an option exists, nor would they know how to do so. Even if they manage it, the last line of `BinaryProvider`'s output clearly states:
```
Info: Could not locate libscsindir inside /home/user/projects/usr/lib
```
just before throwing an error. This is both incorrect and makes it harder to spot messages such as:
```
libopenblas64_.so: cannot open shared object file: No such file or directory
```
which is the real reason for failure.

I consider this more of a paper-cut issue, relevant only to _unsupported_ installations, and would understand if you do not wish to include the error message.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -81,17 +81,6 @@ if !custom_library
         # Download and install binaries
         install(dl_info...; prefix=prefix, force=true, verbose=verbose)
     end
-
-    usable(p) = satisfied(p; verbose=verbose, isolate=true)
-    failed = [p.libnames for p in products if !usable(p)]
-    if !isempty(failed)
-        error(
-            "The following dependencies could not be satisfied by binaries " *
-            "from BinaryProvider: \n$failed.\nIn case You are using non-" *
-            "official Julia binaries, try compiling SCS from source.\nMake " *
-            "sure Julia and SCS are linked against the same OpenBLAS library.\n"
-        )
-    end
 end
 
 # Write out a deps.jl file that will contain mappings for our products
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -76,7 +76,18 @@ if !custom_library
         # Download and install binaries
         install(dl_info...; prefix=prefix, force=true, verbose=verbose)
     end
+
+    usable(p) = satisfied(p; verbose=verbose, isolate=true)
+    failed = [p.libnames for p in products if !usable(p)]
+    if !isempty(failed)
+        error(
+            "The following dependencies could not be satisfied by binaries " *
+            "from BinaryProvider: \n$failed.\nIn case You are using non-" *
+            "official Julia binaries, try installing SCS from source.\nMake " *
+            "sure Julia and SCS are linked against the same OpenBLAS library."
+        )
+    end
 end
 
 # Write out a deps.jl file that will contain mappings for our products
-write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
+#write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -76,6 +76,21 @@ if !custom_library
...
+    # might still try. In that case, BinaryProvider (--verbose) fails with
+    # a confusing message "could not locate," instead of "could not dlopen."
+    usable(p) = satisfied(p; verbose=verbose, isolate=true)
+    failed = [p.libnames for p in products if !usable(p)]
+    if !isempty(failed)
+        error(
+            "The following dependencies could not be satisfied by binaries " *
+            "from BinaryProvider: \n$failed.\nIn case You are using non-" *
+            "official Julia binaries, try compiling SCS from source.\nMake " *
+            "sure Julia and SCS are linked against the same OpenBLAS library.\n"
```
odow, line 91: Or suggest that they use the official `jll` ones that are automatically downloaded? We should probably only run `build.jl` if the following is true
https://github.com/jump-dev/SCS.jl/blob/6611207d8759cb4a7a1f8c10bcff606f0ccb471a/src/SCS.jl#L3

So something like the following at the top of `build.jl`:
```Julia
if !(haskey(ENV, "JULIA_SCS_LIBRARY_PATH") || VERSION < v"1.3")
    @info("Installing default SCS binaries")
    exit(0)
end
```

votroto: Maybe You are right, but I do not understand.

I will give you an example to make sure we are on the same page:
I am running on _ArchLinux_. I have installed _Julia_ as a prebuilt package from the package manager _pacman_ (the official package manager of _ArchLinux_, which provides packages in a "rolling distro" style). Similarly, I have installed _OpenBLAS_ from _pacman_. In this case, running `]add SCS` in the _Julia REPL_ will fail because `libopenblas64_.so: cannot open shared object file: No such file or directory`. As I understand it, this is because the library binary `libscsindir` from _BinaryProvider_ is built and linked against an incompatible version of _OpenBLAS_. On the other hand, building `libscsindir` manually works.

Are you saying that the official `jll` binaries can avoid such issues and that they are downloaded before `build.jl` is executed? If so, then I agree with you. That would be a much better solution.

odow: > Are you saying that the official jll binaries of libscs can avoid such issues and that they are downloaded before build.jl is executed?

Yes. The build script should only be run on Julia 1.2 or less, or if you wish to provide a non-default binary via the `JULIA_SCS_LIBRARY_PATH` environment variable.

Arch can be finicky sometimes. Can you try the change I suggested and see if it worked?

votroto:
You were right. This fixes the issue on two separate machines with non-official Julia installations on versions 1.5.0 and 1.4.2.

I am attaching one  test output.
[testing_output.txt](https://github.com/jump-dev/SCS.jl/files/5058478/testing_output.txt)

Two things:

1. The info message is not visible.
2. Should the check be before or after `using`?

odow:
It should go before the `using`.

The info message should be in `deps/build.log`.

votroto:
Done. 

Indeed, the message is in `deps/build.log`.

kalmarek:
Thanks!

PR #196, opened 2020-08-26
Use dict for cone types

joaquimg:


odow:
Ooof. I can see how this would be bad for inference. Can you kick this off again? https://travis-ci.org/github/jump-dev/SCS.jl/jobs/721448812 

For some reason, I don't have permission.

src/geometric.jl
```diff
@@ -31,21 +37,13 @@ function MOI.empty!(model::GeometricConicForm{T}) where T
...
-
-_findfirst(::Type{S}, sets::Tuple) where {S} = _first(S, 1, sets...)
+_findfirst(s::DataType, d::D) where {D<:AbstractDict} = get(d, s, nothing)
 
 function MOI.supports_constraint(
     model::GeometricConicForm,
     ::Type{MOI.VectorAffineFunction{Float64}},
     ::Type{S}) where S <: MOI.AbstractVectorSet
-    return _findfirst(S, model.cone_types) !== nothing
+    return _findfirst(S, model.cone_types_dict) !== nothing
```
odow, line 46: Remove `_find_first` and just use `haskey(model.cone_types_dict, S)`

src/geometric.jl
```diff
@@ -155,7 +153,7 @@ function MOI.copy_to(dest::GeometricConicForm, src::MOI.ModelLike; preprocess =
 
     has_constraints = BitSet()
     for (F, S) in MOI.get(src, MOI.ListOfConstraints())
-        i = _findfirst(S, dest.cone_types)
+        i = _findfirst(S, dest.cone_types_dict)
```
odow, line 156: `i = get(dest.cone_types_dict, S, nothing)`

src/geometric.jl
```diff
@@ -12,7 +13,12 @@ mutable struct GeometricConicForm{T, AT, C} <: MOI.ModelLike
     dual::Vector{T}
     function GeometricConicForm{T, AT}(cone_types) where {T, AT}
         model = new{T, AT, typeof(cone_types)}()
+
         model.cone_types = cone_types
+        model.cone_types_dict = Dict{DataType, Int}()
+        for (i,s) in enumerate(cone_types)
+            model.cone_types_dict[s] = i
+        end
```
odow, line 21:
```Julia
model.cone_types_dict = Dict{DataType, Int}(
    s => i for (i, s) in enumerate(cone_types)
)
```

src/geometric.jl
```diff
@@ -1,5 +1,6 @@
 mutable struct GeometricConicForm{T, AT, C} <: MOI.ModelLike
     cone_types::C
```
blegat, line 2: Remove

src/geometric.jl
```diff
@@ -1,5 +1,6 @@
 mutable struct GeometricConicForm{T, AT, C} <: MOI.ModelLike
```
blegat, line 1: Remove `, C`

src/geometric.jl
```diff
@@ -13,6 +14,9 @@ mutable struct GeometricConicForm{T, AT, C} <: MOI.ModelLike
     function GeometricConicForm{T, AT}(cone_types) where {T, AT}
         model = new{T, AT, typeof(cone_types)}()
         model.cone_types = cone_types
```
blegat, line 16: Remove line

blegat:
What is the issue exactly ? It works well for inference when `cone_types` is a `Tuple`. I was hoping that the compiler would optimize it out with constant extrapolation but it is not the case so it's O(k) when there is k cones hence using Dict makes sense.
@joaquimg Would still be interested by a comment on where this PR comes from :) Did you identify a bottleneck in the benchmarks ?

odow:
The functions were being invalidated/not compiled properly: https://github.com/jump-dev/MOIPaperBenchmark/pull/11#issuecomment-680992599.

PR #197, opened 2020-10-08
Initial work supporting SCS 3

mlubin:
This is very early work to get the wrapper working with SCS 3. The new version introduces quadratic objectives and the "box" cone. I compiled and tested locally against SCS commit `50900c31106c6dc289a4ef700a148a01cf0c1c9d`. It's obviously premature to merge this before SCS 3 is released. I'm posting it primarily to avoid someone duplicating the work in the future.

@bodono 

TODO:
- [ ] Delete MPB interface (in a separate PR)

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -256,10 +256,14 @@ function MOI.optimize!(optimizer::Optimizer)
     end
 
     linear_solver, options = sanitize_SCS_options(options)
+    P = spzeros(n, n)
 
     cone = optimizer.cone
-    sol = SCS_solve(linear_solver, m, n, _managed_matrix(data.A), b, c,
-                    data.num_rows[1], data.num_rows[2], cone.qa, cone.sa, div(data.num_rows[5], 3), div(data.num_rows[6], 3), cone.p,
+    sol = SCS_solve(linear_solver, m, n, _managed_matrix(data.A),
+                    ManagedSCSMatrix{Int64}(n, n, P), b, c,
+                    data.num_rows[1], data.num_rows[2],
+                    Float64[], Float64[],
+                    cone.qa, cone.sa, div(data.num_rows[5], 3), div(data.num_rows[6], 3), cone.p,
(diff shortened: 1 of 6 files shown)


kalmarek:
@mlubin @bodono any timeline on SCS-3.0.0? (also: is there a chance for the direct GPU solver that was disabled;)
The GPU part with CUDA_jll is almost ready and was planned as SCS-0.8.0, but should we maybe wait for SCS-3.0.0?

bodono:
I don't have a timeline unfortunately, I'm just developing this in my spare time. My guess is that Jan 2020 is realistic for 3.0.0 to land in master. That being said I would say the 3.0.0 api is _mostly_ stable at this point, most of the changes from now will be under the hood. It's in the `qp2` branch (though currently undocumented I'm afraid). I haven't updated the gpu indirect solver at all yet though.

As for the direct GPU solver, I never did get around to fully implementing it. I did have a very simple idea to at least get some idea for the speedup it might offer. The idea was to perform the factorization on cpu using the code that's already there, then ship the factors over to the gpu and use the gpu to do the solves. CuSparse has the *sv2 set of routines that perform sparse triangular solves (https://docs.nvidia.com/cuda/cusparse/index.html#level2-reference), which they claim are quite efficient. One day I'd like to get around to it, or if anyone else would like to that would be great!

kalmarek:
@bodono thanks for the answer, we'll proceed with SCS.jl-0.8.0 as planned then!

kalmarek:
@mlubin I opened #211 because I had forgotten this one existed; I'm fine with abandoning it if you want to carry on here. If not, I can continue #211 (probably squash+rebase onto soon to be master), but I won't be able to spend much time in the next week or two.

mlubin:
I don't have any short-term plans to pick this up (the branch is rather old now). Please continue with #211 :)

bodono:
Sorry @kalmarek , I knew about #211 but somehow forgot about it when I started opening issues!

odow:
Closing in favor of #211

PR #198, opened 2020-10-15
Add GPU support

odow:
This PR is a rebase of #187 onto the `jump-dev` version of SCS.jl so we can run the JuliaGPU tests.

I followed the instructions here: https://github.com/JuliaGPU/gitlab-ci, and I think I have things set up correctly. I added a check for `GITLAB_CI` so that we only run the GPU tests on JuliaGPU, and default to the main tests on Travis.

Closes #187


odow:
@kalmarek ideas? https://gitlab.com/JuliaGPU/SCS-jl/-/jobs/791684624#L169

test/test_gpu.jl
```diff
@@ -1,6 +1,10 @@
 # This file requires a GPU in order to run. It gets tested as part of the
 # JuliaGPU CI on Gitlab. Contact @odow for more details.
 
+# CUDA_jll is not in our Project.toml, so we need to install it on GITLAB_CI.
+import Pkg
+Pkg.add("CUDA_jll")
```
kalmarek, line 6:
```suggestion
pkg"add CUDA_jll@9.0"
```
I actually don't known how to use `PackageSpec` with `Pkg.add` to achieve this ;)
`SCS_GPU` was built against this version:

https://github.com/JuliaPackaging/Yggdrasil/blob/01037ee342dd40ea57c3ad20e69ad8bc81d24bbe/S/SCS_GPU/build_tarballs.jl#L38

This was an arbitrary choice, I picked the lowest available to support possibly widest range of GPUs

kalmarek (commented):
hopefully this will do the trick

odow:
Now we have the fun error:

> ERROR: LoadError: LoadError: ArgumentError: Package Pkg not found in current path:
>  Run `import Pkg; Pkg.add("Pkg")` to install the Pkg package.

https://gitlab.com/JuliaGPU/SCS-jl/-/jobs/793855432#L276

kalmarek:
It's marvelously recurrent :D tbh I don't know what to do about this, but I asked on slack ;)

README.md
```diff
@@ -77,7 +78,32 @@ optimize!(problem)
...
+julia> using SCS
+
+julia> SCS.available_solvers
+2-element Array{DataType,1}:
+ SCS.DirectSolver
+ SCS.IndirectSolver
+```
+
+```julia
+julia> using CUDA_jll # optional, to enable SCS.GpuIndirectSolver
```
odow, line 96: TODO: add instructions on how to add the correct version of CUDA_jll

odow:
Success! I didn't setup codecov yet. It looks like I need to give Gitlab some secret. But we're passing tests at least, and the linear system solves are on a sparse-indirect GPU:
https://gitlab.com/JuliaGPU/SCS-jl/-/jobs/794254537#L3441

kalmarek:
that is great! I didn't notice that we had separate `[test]` category.

As for tests: maybe we could turn `for T in solvers` loop in `test/MOI_wrapper.jl` into a function of `T` and just call it here with `SCS.GpuIndirectSolver`?

kalmarek:
go or no go?

Is it breaking enough to justify major version bump? technically speaking no officially supported parts were altered, but we require new packages. I'm fine eitherway

odow:
Should be fine as a patch release. As you say, this exclusively adds new features and doesn't break any existing code.

kalmarek:
when CI is green, we're good to go and merge this ;)

PR #205, opened 2021-03-06
Replace Inf with a large number in constraints

### LINKED ISSUE #170: SCS outputs NaNs when given non-finite bounds
```julia
using JuMP, SCS
m = Model(with_optimizer(SCS.Optimizer))
@variable(m, -Inf <= x <= Inf)
@objective(m, Max, x)
@constraint(m, x <= 10)
optimize!(m)
```
```
----------------------------------------------------------------------------
 Iter | pri res | dua res | rel gap | pri obj | dua obj | kap/tau | time (s)
----------------------------------------------------------------------------
     0|     -nan       nan      -nan       nan       nan       nan  3.54e-05 
   100|     -nan       nan      -nan       nan       nan       nan  1.41e-04 
   200|     -nan       nan      -nan       nan       nan       nan  2.30e-04 
...
  5000|     -nan       nan      -nan       nan       nan       nan  4.51e-03 
----------------------------------------------------------------------------
Status: Unbounded/Inaccurate
Hit max_iters, solution may be inaccurate
Timing: Solve time: 4.52e-03s
	Lin-sys: avg # CG iterations: 1.00, avg solve time: 1.32e-07s
	Cones: avg projection time: 1.86e-08s
	Acceleration: avg step time: 5.05e-07s
----------------------------------------------------------------------------
Certificate of dual infeasibility:
dist(s, K) = 0.0000e+00
|Ax + s|_2 * |c|_2 = nan
c'x = nan
============================================================================
```
This is related to an issue that @ccoffrin opened but I couldn't find it easily.

At a minimum, the MOI wrapper should throw an error in this case. It would be nicer if the non-finite bounds were just dropped.

odow:
This seemed easier than not-adding the constraint to the model and having to keep track of it.

Closes #170

### DIFF
--- a/src/SCS.jl
+++ b/src/SCS.jl
@@ -31,6 +31,8 @@ include("types.jl")
 
 const available_solvers = [DirectSolver, IndirectSolver]
 
+const SCS_INFINITY = 1e100
+
 include("c_wrapper.jl")
 include("MPB_wrapper.jl")
 include("MOI_wrapper.jl")
--- a/src/geometric.jl
+++ b/src/geometric.jl
@@ -134,6 +134,13 @@ function _load_constraints(model::GeometricConicForm, src, indexmap, cone_offset
         new_func = preprocess(func, set)
         load_terms(model.A, indexmap, new_func, offset)
         copyto!(model.b, offset + 1, new_func.constants)
+        for i = 1:length(new_func.constants)
+            if new_func.constants[i] > SCS_INFINITY
+                model.b[offset + i] = SCS_INFINITY
+            elseif new_func.constants[i] < -SCS_INFINITY
+                model.b[offset + i] = -SCS_INFINITY
+            end
+        end
         model.dimension[offset] = MOI.output_dimension(func)
         indexmap[ci_src] = typeof(ci_src)(offset)
     end
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -67,3 +67,17 @@ end
     @test MOI.get(model, MOI.RawParameter(:eps)) == 2.0
     @test MOI.get(model, MOI.RawParameter("eps")) == 2.0
 end
+
+@testset "Inf bounds" begin
+    model = MOI.Bridges.full_bridge_optimizer(
+        MOI.Utilities.CachingOptimizer(
+            MOI.Utilities.Model{Float64}(),
+            SCS.Optimizer(),
+        ),
+        Float64,
+    )
+    x = MOI.add_variable(model)
+    MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(Inf))
+    MOI.optimize!(model)
+    @test !isnan(MOI.get(model, MOI.VariablePrimal(), x)
+end
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -79,5 +79,5 @@ end
     x = MOI.add_variable(model)
     MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(Inf))
     MOI.optimize!(model)
-    @test !isnan(MOI.get(model, MOI.VariablePrimal(), x)
+    @test !isnan(MOI.get(model, MOI.VariablePrimal(), x))
 end


kalmarek:
I'm not sure it actually fixes the problem. We don't get nans in the problem, but with 
```julia
function test_inf_problem(bound, eps=1e-5)
    m = Model(optimizer_with_attributes(SCS.Optimizer, "eps"=>eps))
    @variable(m, -bound <= x <= bound)
    @objective(m, Max, x)
    @constraint(m, x <= 10)
    optimize!(m)
end
```
(taken from #170), this branch fails with `(bound=Inf, eps=1e-2)` (and all smaller eps).
With `(bound ≥ 1e13, eps=1e-5)` I get
```
----------------------------------------------------------------------------
        SCS v2.1.2 - Splitting Conic Solver
        (c) Brendan O'Donoghue, Stanford University, 2012
----------------------------------------------------------------------------
Lin-sys: sparse-indirect, nnz in A = 3, CG tol ~ 1/iter^(2.00)
eps = 1.00e-05, alpha = 1.50, max_iters = 5000, normalize = 1, scale = 1.00
acceleration_lookback = 10, rho_x = 1.00e-03
Variables n = 1, constraints m = 3
Cones:  linear vars: 3
Setup time: 1.49e-04s
----------------------------------------------------------------------------
 Iter | pri res | dua res | rel gap | pri obj | dua obj | kap/tau | time (s)
----------------------------------------------------------------------------
     0| 1.11e+19  1.82e+18  1.00e+00 -1.99e+31 -3.65e+19  1.04e+32  4.47e-05 
   100| 1.57e-17  0.00e+00  2.31e-05 -1.00e+01 -1.00e+01  9.05e-05  5.93e-04 
   200| 1.57e-17  0.00e+00  2.31e-05 -1.00e+01 -1.00e+01  9.05e-05  1.07e-03
 ...
  4800| 1.57e-17  0.00e+00  2.31e-05 -1.00e+01 -1.00e+01  9.05e-05  2.76e-02 
  4900| 1.57e-17  0.00e+00  2.31e-05 -1.00e+01 -1.00e+01  9.05e-05  2.81e-02 
  5000| 1.57e-17  0.00e+00  2.31e-05 -1.00e+01 -1.00e+01  9.05e-05  2.87e-02 
----------------------------------------------------------------------------
Status: Solved/Inaccurate
Hit max_iters, solution may be inaccurate, returning best found solution.
Timing: Solve time: 2.87e-02s
        Lin-sys: avg # CG iterations: 0.00, avg solve time: 2.81e-07s
        Cones: avg projection time: 7.73e-08s
        Acceleration: avg step time: 4.17e-06s
----------------------------------------------------------------------------
Error metrics:
dist(s, K) = 7.0755e-04, dist(y, K*) = 0.0000e+00, s'y/|s||y| = -5.0031e-17
primal res: |Ax + s - b|_2 / (1 + |b|_2) = 1.5746e-17
dual res:   |A'y + c|_2 / (1 + |c|_2) = 0.0000e+00
rel gap:    |c'x + b'y| / (1 + |c'x| + |b'y|) = 2.3089e-05
----------------------------------------------------------------------------
c'x = -10.0005, -b'y = -10.0000
============================================================================
```

@bodono Maybe there's a chance for a fix on scs level?

odow:
Maybe we should drop those constraints then

bodono:
There might be an issue here with the indirect method that I can look into, I think the direct solver should be the default solver used everywhere unless the problem is so large that only the indirect solver can handle it. Does that setting need to be updated somewhere?

On the larger point - infinite values should be dropped rather than replaced with large numbers in SCS 1 and 2, since the large values make the problem very poorly conditioned which makes the solver struggle (same with other solvers that target the same formulation). In SCS 3 I have added a 'box cone', which can accept infinite / very large values without issue.

kalmarek:
@odow shall we take the opportunity and change the default linear_solver of SCS.jl to `DirectSolver`?
I remember setting it to `IndirectSolver` (`src/types.jl:230`), but frankly speaking I don't remember the rationale for doing so.

blegat:
The constraints should be dropped in the bridge. Infinite bounds are accepted in `Interval` constraints, not sure if we want to accept it in GreaterThan/LessThan constraints, and infinite values are not accepted in functions. So we should fix the `SplitInterval` bridge so that it does create a constraint in case the bound is infinite.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/issues/1270.

odow:
> shall we take the opportunity and change the default linear_solver of SCS.jl to DirectSolver?

Yes.

PR #206, opened 2021-03-08
make DirectSolver the default one

kalmarek:
> I think the direct solver should be the default solver used everywhere unless the problem is so large that only the indirect solver can handle it.

https://github.com/jump-dev/SCS.jl/pull/205#issuecomment-791938069

### DIFF
--- a/test/options.jl
+++ b/test/options.jl
@@ -21,23 +21,22 @@ m = MathProgBase.ConicModel(s)
 MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
 MathProgBase.optimize!(m)
 
-@test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-9, rtol=0.0)
-@test !isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-10, rtol=0.0)
+@test isapprox(MathProgBase.getobjval(m), -99.0, rtol=1e-9)
+@test !isapprox(MathProgBase.getobjval(m), -99.0, rtol=1e-10)
+# we test above to check if the next optimize actually does something
 
 # With eps = 1e-14, solution should be far more accurate
 s = SCSSolver(eps=1e-14)
 m = MathProgBase.ConicModel(s)
 MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
 MathProgBase.optimize!(m)
-@test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-12, rtol=0.0)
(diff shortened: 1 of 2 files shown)


odow:
See the failing tests. Seems like some tolerance issues.

test/options.jl
```diff
@@ -49,7 +49,7 @@ MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
 MathProgBase.setwarmstart!(m, primal_sol; dual_sol = dual_sol, slack = slack)
 MathProgBase.optimize!(m)
 @test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-12, rtol=0.0)
-@test !isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-14, rtol=0.0)
+@test !isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-13, rtol=0.0)
```
odow, line 52: I think you're going to have to make this waaaay looser. Like `atol=1e-6`.

kalmarek: heh, It seems to me that I need to actually tighten this. all of tests fail with
```
Test Failed at /Users/runner/work/SCS.jl/SCS.jl/test/options.jl:52
  Expression: !(isapprox(MathProgBase.getobjval(m), -99.0, atol = 1.0e-13, rtol = 0.0))
```
i.e. solution is even more accurate ;)

kalmarek: ok, I decided to comment those two lines as asking for `atol=1e-15` or smaller seems silly

odow: Urgh. I didn't see the `!` at the start of the line. That's a weird test. Why are we checking that we didn't find an accurate solution? Commenting seems fine.

test/options.jl
```diff
@@ -21,8 +21,8 @@ m = MathProgBase.ConicModel(s)
 MathProgBase.loadproblem!(m, -obj, A, rowub, [(:NonNeg,1:3)],[(:NonNeg,1:5)])
 MathProgBase.optimize!(m)
 
-@test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-9, rtol=0.0)
-@test !isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-10, rtol=0.0)
+@test isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-7, rtol=0.0)
+@test !isapprox(MathProgBase.getobjval(m), -99.0, atol=1e-8, rtol=0.0)
```
odow, line 25: How about removing this one as well?

odow (approved):
Merge if you're happy.

test/options.jl
kalmarek: Very well! I left only one lower bounds test with a comment.

kalmarek:
sorry, I completely forgot about this one!

PR #210, opened 2021-05-24
Conic form supports variable primal

blegat:
We're calling `copy_to` to this `GeometricConicForm` directly:
https://github.com/jump-dev/SCS.jl/blob/7669c8df10a71da367e6bf7fe37a3ee7aba144a1/src/MOI_wrapper.jl#L146
 so it needs to implement `supports` otherwise the starting values are not copied and we get the warning
```
┌ Warning: MathOptInterface.VariablePrimalStart() is not supported by SCS.GeometricConicForm{Float64, SCS.SparseMatrixCSRtoCSC{Int64}, NTuple{8, DataType}}. This information will be discarded.
```

### DIFF
--- a/src/geometric.jl
+++ b/src/geometric.jl
@@ -54,6 +54,13 @@ function rows(model::GeometricConicForm, ci::CI{MOI.VectorAffineFunction{Float64
     return ci.value .+ (1:model.dimension[ci.value])
 end
 
+function MOI.supports(
+    ::GeometricConicForm,
+    ::MOI.VariablePrimalStart,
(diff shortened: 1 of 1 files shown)


kalmarek:
@blegat I think scs library will disregard primal warmstart unless the compatible dual and slack are provided as well. Since we fill those with zeros, I don't think setting just the primal will have real effect (but I haven't actually checked it!).

Have you tried using it? does it work?

blegat:
No I haven't checked whether SCS was behaving differently

blegat:
Added constraint and dual starts

PR #219, opened 2021-09-14
Breaking changes for MOI 0.10

odow:
More test failures than we should be happy with. I'll comment in-line.


test/MOI_wrapper.jl
```diff
@@ -1,73 +1,117 @@
...
+            # Unexpected test failures:
+            #   Failures getting ConstraintFunction
+            "test_basic_ScalarQuadraticFunction_LessThan",
+            "test_basic_VectorOfVariables_GeometricMeanCone",
+            "test_conic_SecondOrderCone_negative_initial_bound",
+            "test_model_ModelFilter_AbstractConstraintAttribute",
+            "test_model_copy_to_UnsupportedAttribute",
+            "test_modification_set_function_single_variable",
+            "test_objective_incorrect_modifications",
+            "test_solve_optimize_twice",
```
odow, line 62: Plenty of failures here to debug. A lot of these look upstream. For example the solve objective twice one is due to our `copy_to_and_optimize!`. Obviously something isn't being set right in the caching optimizer.

test/MOI_wrapper.jl
```diff
@@ -1,73 +1,117 @@
...
+            #   cone dimensions 241 not equal to num rows in A = m = 31
+            "test_conic_LogDetConeTriangle",
+            "test_conic_NormNuclearCone",
+            "test_conic_NormSpectralCone",
+            "test_conic_PositiveSemidefiniteConeSquare_3",
+            "test_conic_PositiveSemidefiniteConeSquare_VectorAffineFunction",
+            "test_conic_PositiveSemidefiniteConeSquare_VectorOfVariables",
+            "test_conic_PositiveSemidefiniteConeTriangle",
+            "test_conic_RootDetConeTriangle",
```
odow, line 71: @blegat These are all very weird ones. We're obviously not passing our PSD matrices correctly?

test/MOI_wrapper.jl
```diff
@@ -1,73 +1,117 @@
...
+            "test_basic_ScalarQuadraticFunction_GreaterThan",
+            "test_basic_ScalarQuadraticFunction_Interval",
+            "test_basic_VectorQuadraticFunction_",
+            "test_quadratic_SecondOrderCone_basic",
+            "test_quadratic_nonconvex_",
+            #   ConstraintSet not implemented
+            "test_basic_VectorAffineFunction_LogDetConeTriangle",
+            "test_basic_VectorAffineFunction_RootDetConeTriangle",
+            "test_basic_VectorOfVariables_LogDetConeTriangle",
+            "test_basic_VectorOfVariables_RootDetConeTriangle",
```
odow, line 91: Features missing in upstream bridges

test/MOI_wrapper.jl
```diff
@@ -1,73 +1,117 @@
...
+            #   power cone error, values must be in [-1,1]
+            "test_conic_DualPowerCone_VectorOfVariables",
+            "test_conic_DualPowerCone_VectorAffineFunction",
+            "test_conic_PowerCone_VectorAffineFunction",
+            "test_conic_PowerCone_VectorOfVariables",
```
odow, line 96: Okay this is a pain. SCS supports a modified version of the power cone.

test/MOI_wrapper.jl
```diff
@@ -1,73 +1,117 @@
...
+            # TODO(odow): fix upstream
+            "test_model_ListOfConstraintAttributesSet",
+            "test_objective_FEASIBILITY_SENSE_clears_objective",
```
odow, line 99: I think these are my problems upstream.

test/MOI_wrapper.jl
odow: This was a double-header. We were passing the sets incorrectly: https://github.com/jump-dev/MathOptInterface.jl/pull/1607, but we're also not scaling the off-diagonal PSD values by `sqrt(2)`.

odow: @blegat are there any tools for working with permuting these matrices?

We need to:
 * Convert from upper-triangular to lower-triangular
 * Scale the off-diagonal entries by `sqrt(2)`
 * Scale the corresponding solution entries by `1 / sqrt(2)`

I see there is a lot of code to do this in the old MOI wrapper, but you didn't have any in your new PR.

odow: I've adding the scaling. I just need to do the permutation of upper-triangular to lower-triangular

odow:
Okay, after this experience, I think I'm now on board with:
 1) The required bridges thing. It made SCS much simpler than the direction I was originally going down. 
 2) The `OptimizerCache` stuff. Not necessarily as a replacement layer, but as a simpler way of converting into matrix-type formats
3) The one-shot `optimize!(dest, src)`.

All-in-all, this rewrite of SCS.jl was not a fun experience, but I think we're in a much better place than we were.

odow:
@blegat did you actually try https://github.com/jump-dev/MathOptInterface.jl/pull/1592 on a solver? It isn't working.

odow:
`empty!` is not reaching the inner optimizer
```Julia
julia> model = MOI.Bridges.full_bridge_optimizer(
           MOI.Utilities.CachingOptimizer(
               MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
               SCS.Optimizer(),
           ),
           Float64,
       );

julia> MOI.set(model, MOI.Silent(), true)

julia> x, ci = MOI.add_constrained_variables(model, MOI.Nonnegatives(2));

julia> MOI.optimize!(model)

julia> MOI.empty!(model)

julia> MOI.get(model, MOI.TerminationStatus())
OPTIMAL::TerminationStatusCode = 1
```

PR #225, opened 2021-09-20
Fix VectorOfVariables-in-PositiveSemidefiniteConeTriangle

odow:
Closes #224 

The `SetMap` bridges are set up in MOI such that the bridge should accept any incoming function type. That means that this wasn't getting `functionized` in between.

I looked at changing things on the MOI end, but it was easier to just add support here.

PR #226, opened 2021-09-20
Fix ScaledPSDConeBridge for VectorOfVariables

blegat:
See https://github.com/jump-dev/SCS.jl/issues/224


odow:
I guess we should assign issues, instead of both working on the same thing

blegat:
I agree, we should rely more on github assignment. Since you opened the issue without assigning yourself, I didn't expect you were working on it so I assigned myself and worked on it. Now that I have done this PR, I have unassigned myself as I'm not working on adding the test to MOI at the moment.

odow:
Closing in favor of #225

PR #229, opened 2021-10-14
Throw UnsupportedConstraint

blegat:
cc @kalmarek


kalmarek (approved):
I don't know enough about the MOI wrapper to have an opinion; @odow merge at will

PR #242, opened 2022-01-23
Only pass warm-start if explicitly set

### LINKED ISSUE #241: SCS v0.8.2 breaks JuMP documentation
x-ref: https://github.com/jump-dev/JuMP.jl/pull/2857

odow:
We should only tell SCS to warm-start the solve if the user has explicitly passed one.

Closes #241

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -245,11 +245,13 @@ function MOI.optimize!(
         fill!(dest.sol.slack, 0.0)
     end
     # Set starting values and throw error for other variable attributes
+    has_warm_start = false
     vis_src = MOI.get(src, MOI.ListOfVariableIndices())
     for attr in MOI.get(src, MOI.ListOfVariableAttributesSet())
         if attr == MOI.VariableName()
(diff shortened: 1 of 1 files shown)


kalmarek:
this looks good to me; 

Side comment: scs will ignore provided warmstart if not all `(variable_primal, constraint_primal, constraint_dual)` are passed. Should we warn user about this behavior?
if so, one could turn `has_warm_start` into a tuple `has_warm_start = (variable_primal=false, constraint_primal=false, constraint_dual=false)`, emit a warning if `any(has_warm_start) && !all(has_warm_start)` and pass the latter to `scs_solve`

odow:
If the user only passes some warm-starts, we fill the rest with zeros so I think we're okay. The problem with the JuMP example was that previously we didn't warm start anything, and now we're warming starting with 0s if the user does nothing.

kalmarek:
we are fine, but the user may be surprised that despite passing primal warmstart e.g. solver starts far away ;)

Anyway, I don't really know what are the conventions of warnings in JuMP, so just wanted to bring this to your attention

kalmarek (approved):

PR #251, opened 2022-02-28
Mk/scs 320

kalmarek:
https://github.com/JuliaPackaging/Yggdrasil/pull/4500 build work with this, locally

### DIFF
--- a/src/c_wrapper.jl
+++ b/src/c_wrapper.jl
@@ -310,9 +310,7 @@ function scs_solve(
     Avalues, Arowval, Acolptr = _to_sparse(T, A)
     Pvalues, Prowval, Pcolptr = _to_sparse(T, P)
     option_dict = _sanitize_options(options)
-    if warm_start
-        option_dict[:warm_start] = 1
-    end
+    option_dict[:warm_start] = convert(Int, warm_start)
     model = _ScsDataWrapper(
         linear_solver,
         m,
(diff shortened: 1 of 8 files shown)


src/c_wrapper.jl
```diff
@@ -74,6 +74,7 @@ end
 mutable struct ScsInfo{T} <: AbstractSCSType
     iter::T
     status::NTuple{128,Cchar}
+    lin_sys_solver::NTuple{128,Cchar}
```
odow, line 77: :( This isn't `3.2` then! That's a breaking change...

src/linear_solvers/direct.jl
```diff
@@ -24,16 +24,31 @@ function scs_init(
...
+        c::Ref{Float64},
+    )::Clonglong
+end
+
 function scs_solve(
     ::Type{DirectSolver},
     work::Ptr{Cvoid},
     solution::ScsSolution,
     info::ScsInfo{I},
+    warm_start::Integer,
```
odow, line 45: :( More breaking changes

src/c_wrapper.jl
kalmarek: only if `ScsInfo` is part of the official interface. Since we don't export it I don't consider it as such.

src/linear_solvers/direct.jl
kalmarek: this seems to be breaking, but we can fix it by giving it the default value `0`.

kalmarek: Also: we do export `scs_solve`, but this particular method isn't even documented (e.g. because of its arguments which are private), so I think these changes justify only minor version bump (but maybe we should opt for `1.1.0`?

src/c_wrapper.jl
odow: I meant in the underlying SCS library

src/linear_solvers/direct.jl
odow: Again, I meant in the underlying library. I think our only public API is the MOI optimizer and the full `SCS.scs_solve`.

src/c_wrapper.jl
kalmarek: ok, then I misunderstood :)

PR #253, opened 2022-03-17
fix passing strings options to ScsSettings

kalmarek:


odow (commented):
How's this? I added a `GC.@preserve` to make sure the result of `cconvert` doesn't get freed during the `setproperty!`.

src/c_wrapper.jl
```diff
@@ -370,7 +370,15 @@ function _unsafe_scs_solve(model::_ScsDataWrapper{S,T}) where {S,T}
         pointer(model.c),
     )
     for (key, value) in model.options
-        setproperty!(model.settings, key, value)
+        if value isa AbstractString
+            c_value = Base.cconvert(Cstring, value)
+            unsafe_value = Base.unsafe_convert(Cstring, c_value)
+            GC.@preserve c_value begin
```
kalmarek, line 376: yeah, in general supporting something that will be non-trivially `cconvert`ed is hard. Suppose that I have `a::MyFancyString` such that `Base.cconvert(Cstring, a)` = String(a.some_data)` (newly constructed string). Then `unsafe_value = Base.unsafe_convert(Cstring,  c_value)` is basically a pointer to `c_value` and while `c_value` is preserved during `setproperty!` we will be left with freed pointer in `model.settings` after GC kicks.

We'd need to make sure that `model.options` holds data that will _not_ be non-trivially `cconvert`ed...

kalmarek: just checked and for _anything but_ `String` `cconvert` allocates new `String` ;)

how about checking that `c_value === value` and throwing asking user to provide options with honest `Strings`?

odow: Oh. Yeah I didn't think about the pointer `model.settings` part. How about just
```Julia
if value isa AbstractString && !(value isa String)
    error("You must pass a `String` as the value for $(key)") 
end
```

kalmarek: another option (which I like a bit less) is to replace the value in `model.options` by `Base.cconvert(Cstring, value)`.

src/c_wrapper.jl
```diff
@@ -370,7 +370,14 @@ function _unsafe_scs_solve(model::_ScsDataWrapper{S,T}) where {S,T}
         pointer(model.c),
     )
     for (key, value) in model.options
-        setproperty!(model.settings, key, value)
+        if value isa AbstractString
+            value isa String ||
```
odow, line 374: Please don't use this short-circuiting, prefer the expanded form. It means that coverage isn't accurate, and we don't end up testing the `|| other` case.

kalmarek: sure, if this is discouraged, then maybe it should be added as option to formatting? see e.g. https://domluna.github.io/JuliaFormatter.jl/stable/api/#JuliaFormatter.is_standalone_shortcircuit-Tuple{CSTParser.EXPR}

odow: Looks like there's a function, but it's not exposed as a setting: https://domluna.github.io/JuliaFormatter.jl/stable/api/#JuliaFormatter.format_text-Tuple{AbstractString}

odow (approved):

PR #264, opened 2023-04-03
Fix to get tests passing

odow:


blegat:
What's the issue with that test ?

blegat (approved):

PR #267, opened 2023-04-07
update to SCS-3.2.3

kalmarek:


odow:
@kalmarek GPU is failing with 
image

do we need to compile with `-DINDIRECT=1` or something?

odow:
Previously we were segfaulting
image
so that's progress. 

@bodono's release notes suggest that the GPU should now work:
image

Project.toml
```diff
@@ -14,9 +14,9 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 [compat]
 MathOptInterface = "1"
 Requires = "1"
-SCS_GPU_jll = "=3.2.1"
-SCS_MKL_jll = "=3.2.2"
-SCS_jll = "=3.2.1"
+SCS_GPU_jll = "=3.2.1, =3.2.3"
```
odow, line 17:
```suggestion
SCS_GPU_jll = "=3.2.3"
```

odow:
GPU build can come separately.

PR #270, opened 2023-05-22
Change ScaledPSDConeBridge to support ScaledPositiveSemidefiniteConeTriangle

blegat:


src/MOI_wrapper/scaled_psd_cone_bridge.jl
```diff
@@ -3,6 +3,14 @@
...
 
+"""
+    struct ScaledPSDCone <: MOI.AbstractVectorSet
+        side_dimension::Int
+    end
+
+Similar to `MOI.ScaledPositiveSemidefiniteConeTriangle` but it the vectorization
+is the lower triangular part column-wise (or the upper triangular part row-wise).
+"""
 struct ScaledPSDCone <: MOI.AbstractVectorSet
```
odow, line 14: This seems a bit pointless if we're still going to need the bridge...

Now `@constraint(model, X in PSDCone())` is going to need two bridges instead of one...

blegat: What's the issue of using several bridges ? We should go for simple atomic bridges. If there is an issue about intermediate copies, we can think about fusing the operations or mutating the functions instead of copying them but I'd rather keep the solver wrappers simple and let MOI do the tricky stuff in view of the growing list of solvers.

blegat: Also, this allows passing scaled PSD constraints to SCS directly while without this PR it would unscale and then scale again

odow: There's nothing wrong with multiple bridges. (Except that it'll make https://jump.dev/JuMP.jl/stable/tutorials/conic/ellipse_approx/#Alternative-formulations more verbose.)

But it just seems pointless to rescale the matrix only to then permute it in a different bridge. With the new test system, the overhead for creating and maintaining a bridge is pretty low and this one in particular is well tested.

Perhaps we could have two bridges in SCS?

odow: > By the way, the PSD cone is missing from the README

Yes. I just realized that. It's because it needs a bridge and we decided that solvers shouldn't report all sets that they support with bridges :smile: But we could add `SCS.ScaledPSDCone`

blegat: > But it just seems pointless to rescale the matrix only to then permute it in a different bridge.

I don't see why that's pointless. The bridges are really just matrix multiplication, here we are multiplying by a diagonal and a permutation matrix, that makes sense that it's the product of two different matrices, also it makes it apparent which one is multiplied first since they are not commutative.

> With the new test system, the overhead for creating and maintaining a bridge is pretty low and this one in particular is well tested.

I'd rather not have both. Note that there is also the issue of allocating this scaling vector that was resolved in the MOI bridge (see https://github.com/jump-dev/SCS.jl/issues/228). This was reported in 2019 in https://github.com/jump-dev/SCS.jl/pull/165 but not reported again. That doesn't mean that no user gets any issue, we just gets a very small percentage of issues people encounter as bug report I'd still rather have fewer bridges that are hit by more users.

> But we could add `SCS.ScaledPSDCone`

Yes, I think we should :-P

odow: The performance problem was a bug. It just used a poor choice of data structure.

blegat: I don't see why a `BitSet` was worse than a `Vector{Float64}` (which is what is used now by the bridge). In any case, that still shows that we probably don't want to duplicate the number of equivalent bridges.

odow: The `BitSet` problem is separate to the `scale` `Vector{Float64}`. 

We used to do:
```julia
diagidx = BitSet()
for i in 1:d
    push!(diagidx, trimap(i, i))
end
for i in 1:length(output)
    if !(rows[i] in diagidx)
        output[i] *= scaling
    end
end
```

#165 changed it to:
```julia
for i in 1:length(output)
    val = 8 * rows[i] + 1
    if isqrt(val)^2 != val
        output[i] *= scaling
    end
end
```

The bridge now does:
```julia
scale_factor = fill(scale, d)
for i in 1:d
    if MOI.Utilities.is_diagonal_vectorized_index(i)
        scale_factor[i] = 1.0
    end
end
```

src/MOI_wrapper/scaled_psd_cone_bridge.jl
```diff
@@ -19,27 +27,27 @@ function MOI.dimension(x::ScaledPSDCone)
...
     ScaledPSDCone,
-    MOI.PositiveSemidefiniteConeTriangle,
-    MOI.VectorAffineFunction{T},
-    G,
+    MOI.ScaledPositiveSemidefiniteConeTriangle,
+    F,
+    F,
 }
-    constraint::MOI.ConstraintIndex{MOI.VectorAffineFunction{T},ScaledPSDCone}
+    constraint::MOI.ConstraintIndex{F,ScaledPSDCone}
```
odow, line 37: If we're going to support arbitrary functions, then we need to add tests for at least `MOI.VectorOfVariables`.

src/MOI_wrapper/scaled_psd_cone_bridge.jl
blegat: Yes, the code sample in the middle is the only one that does not allocate, doesn't it ?
Anyway, the issue with keeping the two bridge is that one of them will receive very low coverage which is also an issue, isn't it ?

src/MOI_wrapper/scaled_psd_cone_bridge.jl
blegat: Fair point

src/MOI_wrapper/scaled_psd_cone_bridge.jl
odow: I guess it's not quite my example. Because the matrix is sparse, not dense. 

If you take the example from https://github.com/jump-dev/SCS.jl/issues/228#issuecomment-984293311, there's no performance issue with `SCS#master`. Almost all the time is spent inside SCS. Where it used to take 5 minutes with the `BitSet`.

blegat: Sure, with that example there is no issue but that does not mean that there is no example with an issue. https://github.com/jump-dev/SCS.jl/pull/165 report an example for which there is an issue.

blegat: Oh, ok I missed that they are the same sorry. Not sure why allocating vectors would be cheaper than `BitSet`, maybe it was because we were not using `sizehint!` but I guess we don't have to spend time on this if we just agree that it's best not to have two bridges in SCS as then one will get no coverage ?
I don't see why having more lines in the output of the bridge debugging is bad. There is actually two transformations going on, we need to scale and we need to reorder so it's quite intuitive to see both.
The users also sees that if he has the scaled matrices, he can remove one of the two bridges while if you have one bridge that does the two you won't see that so it's loosing some transparency aspect of the bridging.

odow (approved):
So (1): I approve adding a bridge like this.

But (2): is this not technically breaking if someone was explicitly using the bridge, rather than a `LazyBridgeOptimizer`?

odow:
- [ ] Benchmark this to see if the extra copy makes a difference

odow:
Did some benchmarking. The extra copy is not a cause for concern, so I'm okay to merge this PR.

# Previous performance benchmark

## Code

```Julia
julia> import MathOptInterface as MOI

julia> import Random

julia> import SCS

julia> import SparseArrays

julia> function main()
           Random.seed!(1234)
           m, n = 20, 800
           d = div(n * (n + 1), 2)
           A, b, c = SparseArrays.sprand(d, m, 0.2), rand(d), rand(m)
           model = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
           x = MOI.add_variables(model, m)
           obj = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(c, x), 0.0)
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           terms = map(zip(SparseArrays.findnz(A)...)) do (i, j, v)
               return MOI.VectorAffineTerm(i, MOI.ScalarAffineTerm(-v, x[j]))
           end
           MOI.add_constraint(
               model, 
               MOI.VectorAffineFunction(terms, b),
               MOI.PositiveSemidefiniteConeTriangle(n),
           )
           scs = MOI.instantiate(SCS.Optimizer; with_bridge_type = Float64)
           MOI.set(scs, MOI.RawOptimizerAttribute("max_iters"), 1)
           MOI.copy_to(scs, model)
           MOI.optimize!(scs)
           return
       end
main (generic function with 1 method)
```

## This PR

```julia
julia> @time main()
------------------------------------------------------------------
	       SCS v3.2.3 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 20, constraints m: 320400
cones: 	  s: psd vars: 320400, ssize: 1
settings: eps_abs: 1.0e-04, eps_rel: 1.0e-04, eps_infeas: 1.0e-07
	  alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
	  max_iters: 1, normalize: 1, rho_x: 1.00e-06
	  acceleration_lookback: 10, acceleration_interval: 10
lin-sys:  sparse-direct-amd-qdldl
	  nnz(A): 1280310, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.32e+00 
     1| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.33e+00 
------------------------------------------------------------------
status:  solved (inaccurate - reached max_iters)
timings: total: 1.33e+00s = setup: 5.91e-01s + solve: 7.41e-01s
	 lin-sys: 1.02e-02s, cones: 6.75e-01s, accel: 2.50e-08s
------------------------------------------------------------------
objective = 1726.040603 (inaccurate)
------------------------------------------------------------------
  1.725265 seconds (3.03 k allocations: 313.449 MiB, 2.24% gc time)

julia> @time main()
------------------------------------------------------------------
	       SCS v3.2.3 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 20, constraints m: 320400
cones: 	  s: psd vars: 320400, ssize: 1
settings: eps_abs: 1.0e-04, eps_rel: 1.0e-04, eps_infeas: 1.0e-07
	  alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
	  max_iters: 1, normalize: 1, rho_x: 1.00e-06
	  acceleration_lookback: 10, acceleration_interval: 10
lin-sys:  sparse-direct-amd-qdldl
	  nnz(A): 1280310, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.31e+00 
     1| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.32e+00 
------------------------------------------------------------------
status:  solved (inaccurate - reached max_iters)
timings: total: 1.32e+00s = setup: 5.86e-01s + solve: 7.37e-01s
	 lin-sys: 1.01e-02s, cones: 6.71e-01s, accel: 2.40e-08s
------------------------------------------------------------------
objective = 1726.040603 (inaccurate)
------------------------------------------------------------------
  1.704311 seconds (3.03 k allocations: 313.449 MiB, 2.92% gc time)
```

## master

```julia
julia> @time main()
------------------------------------------------------------------
	       SCS v3.2.3 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 20, constraints m: 320400
cones: 	  s: psd vars: 320400, ssize: 1
settings: eps_abs: 1.0e-04, eps_rel: 1.0e-04, eps_infeas: 1.0e-07
	  alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
	  max_iters: 1, normalize: 1, rho_x: 1.00e-06
	  acceleration_lookback: 10, acceleration_interval: 10
lin-sys:  sparse-direct-amd-qdldl
	  nnz(A): 1280310, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.30e+00 
     1| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.31e+00 
------------------------------------------------------------------
status:  solved (inaccurate - reached max_iters)
timings: total: 1.32e+00s = setup: 5.87e-01s + solve: 7.29e-01s
	 lin-sys: 8.94e-03s, cones: 6.67e-01s, accel: 3.20e-08s
------------------------------------------------------------------
objective = 1726.040603 (inaccurate)
------------------------------------------------------------------
  1.715707 seconds (3.00 k allocations: 271.931 MiB, 1.15% gc time)

julia> @time main()
------------------------------------------------------------------
	       SCS v3.2.3 - Splitting Conic Solver
	(c) Brendan O'Donoghue, Stanford University, 2012
------------------------------------------------------------------
problem:  variables n: 20, constraints m: 320400
cones: 	  s: psd vars: 320400, ssize: 1
settings: eps_abs: 1.0e-04, eps_rel: 1.0e-04, eps_infeas: 1.0e-07
	  alpha: 1.50, scale: 1.00e-01, adaptive_scale: 1
	  max_iters: 1, normalize: 1, rho_x: 1.00e-06
	  acceleration_lookback: 10, acceleration_interval: 10
lin-sys:  sparse-direct-amd-qdldl
	  nnz(A): 1280310, nnz(P): 0
------------------------------------------------------------------
 iter | pri res | dua res |   gap   |   obj   |  scale  | time (s)
------------------------------------------------------------------
     0| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.31e+00 
     1| 1.01e+01  4.69e+02  3.45e+03  1.73e+03  1.00e-01  1.32e+00 
------------------------------------------------------------------
status:  solved (inaccurate - reached max_iters)
timings: total: 1.32e+00s = setup: 5.92e-01s + solve: 7.31e-01s
	 lin-sys: 1.02e-02s, cones: 6.65e-01s, accel: 3.20e-08s
------------------------------------------------------------------
objective = 1726.040603 (inaccurate)
------------------------------------------------------------------
  1.777191 seconds (3.00 k allocations: 271.931 MiB, 2.11% gc time)
```


# Worst-case: lots of little PSD cones

## Code

```julia
julia> import MathOptInterface as MOI

julia> import SCS

julia> function main(N)
           model = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
           obj = zero(MOI.ScalarAffineFunction{Float64})
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           for i in 1:N
               x = MOI.add_variables(model, 6)
               obj += MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(1.0, x), 0.0)
               MOI.add_constraint(
                   model, 
                   MOI.Utilities.operate(vcat, Float64, 1.0 .* x...),
                   MOI.PositiveSemidefiniteConeTriangle(6),
               )
           end
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj)}(), obj)
           scs = MOI.instantiate(SCS.Optimizer; with_bridge_type = Float64)
           MOI.set(scs, MOI.Silent(), true)
           MOI.set(scs, MOI.RawOptimizerAttribute("max_iters"), 1)
           MOI.copy_to(scs, model)
           MOI.optimize!(scs)
           return
       end
main (generic function with 1 method)
```

## This PR

```Julia
julia> for N in (1, 10, 100, 1_000, 10_000)
           GC.gc()
           @time main(N)
       end
  0.001020 seconds (2.79 k allocations: 123.891 KiB)
  0.001285 seconds (4.46 k allocations: 234.672 KiB)
  0.004006 seconds (20.92 k allocations: 2.664 MiB)
  0.066691 seconds (201.96 k allocations: 104.580 MiB, 9.18% gc time)
  3.231644 seconds (2.02 M allocations: 9.050 GiB, 15.38% gc time)
```

## master

```Julia
julia> for N in (1, 10, 100, 1_000, 10_000)
           GC.gc()
           @time main(N)
       end
  0.000948 seconds (2.76 k allocations: 122.062 KiB)
  0.001238 seconds (4.30 k allocations: 225.844 KiB)
  0.003974 seconds (19.51 k allocations: 2.589 MiB)
  0.065522 seconds (187.94 k allocations: 103.846 MiB, 9.92% gc time)
  3.223676 seconds (1.88 M allocations: 9.043 GiB, 16.75% gc time)
```

odow (approved):

blegat:
Thanks for the benchmark! I hope we get a benchmark showing the opposite at some point so that it justifies doing https://github.com/jump-dev/MathOptInterface.jl/issues/526 :innocent:

PR #275, opened 2023-06-05
Fix flush of stdout after solve

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3402

### DIFF
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -191,7 +191,9 @@ function test_redirect_stdout()
             MOI.add_constraint(model, f, MOI.Nonnegatives(3))
             scs = SCS.Optimizer()
             MOI.optimize!(scs, model)
+            return
         end
+        return
     end
     output = read(filename, String)
     @test occursin("SCS", output)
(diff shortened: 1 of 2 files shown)


test/MOI_wrapper.jl
```diff
@@ -181,6 +181,24 @@ function test_Name_skip()
...
+function test_redirect_stdout()
+    filename = tempname()
+    open(filename, "w") do io
+        redirect_stdout(io) do
+            model = MOI.Utilities.Model{Float64}()
+            x = MOI.add_variables(model, 3)
+            f = MOI.Utilities.operate(vcat, Float64, 1.0 .* x...)
+            MOI.add_constraint(model, f, MOI.Nonnegatives(3))
+            scs = SCS.Optimizer()
+            MOI.optimize!(scs, model)
```
odow, line 193:
```suggestion
            MOI.optimize!(scs, model)
            return
```

test/MOI_wrapper.jl
```diff
@@ -181,6 +181,24 @@ function test_Name_skip()
...
+        end
```
odow, line 194:
```suggestion
        end
        return
```

blegat:
Nice catch, I was also wondering why I wasn't seeing the output in Documenter

PR #281, opened 2023-08-15
Update Project.toml

matbesancon:


matbesancon:
Technically it's a new feature so I upgraded the minor release. If we want a bugfix bump instead let me know

PR #283, opened 2023-09-09
Update to the new MOI scaled set

blegat:
See https://github.com/jump-dev/MathOptInterface.jl/issues/2260


kalmarek:
@blegat no minor version bump + register? something still in the pipeline?

blegat:
No,we can release

PR #284, opened 2023-10-24
Fix [compat] in Project.toml

odow:


Project.toml
```diff
@@ -13,10 +13,13 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 MathOptInterface = "1.20"
+Pkg = "1.6"
 Requires = "1"
 SCS_GPU_jll = "=3.2.3"
 SCS_MKL_jll = "=3.2.2, =3.2.3"
 SCS_jll = "=3.2.1, =3.2.3"
+SparseArrays = "1.6"
+Test = "1.6"
```
odow, line 22:
```suggestion
```

odow:
cc @vchuravy I'm getting `LoadError: empty intersection between SparseArrays@0.0.0 and project compatibility 1.6.0-1`.

It looks like it is happening because we `Pkg.add` stuff to the environment during test:
https://github.com/jump-dev/SCS.jl/blob/1a5cb7d9a080c45677b8bb55cdec1e28963e9ff1/test/runtests.jl#L15-L16

Any ideas?

odow:
Closing as won't-fix (for now).

There is a bug in Pkg.jl that doesn't let compat bounds for stdlibs if `Pkg` operations take place in the tests.

It will be fixed with a new release of 1.6.X (and 1.9.X or 1.10.0), so we can re-visit once that happens.

The downside is that any new releases may need a manual merge to General.

PR #286, opened 2023-11-13
[breaking] refactor linear solver loading to use Pkg extensions

odow:
This is a WIP PR to flesh out what a refactored SCS.jl would look like if we used package extensions.

CI will be broken until at least 

- [x] https://github.com/JuliaPackaging/Yggdrasil/pull/7671
- [x] https://github.com/JuliaPackaging/Yggdrasil/pull/7672
- [x] https://github.com/JuliaPackaging/Yggdrasil/pull/7673


kalmarek:
@odow looks rather good, no more requires, init etc, much cleaner;

I'd suggest however to wait with this until the a new LTS is released which does support extensions. In some sense the current situation, while not optimal just works, and I don't feel like putting more burden of maintaining two incompatible versions of SCS.jl

src/SCS.jl
```diff
@@ -6,45 +6,23 @@
...
-
-                push!(available_solvers, MKLDirectSolver)
-            end
-        end
-    )
-    return
-end
+abstract type LinearSolver end
+
+SCS.is_available(::Type{<:LinearSolver}) = false
```
odow, line 14:
```suggestion
is_available(::Type{<:LinearSolver}) = false
```

ext/SCSSCS_GPU_jllExt.jl
```diff
@@ -0,0 +1,75 @@
...
+
+SCS.scsint_t(::Type{SCS.GpuIndirectSolver}) = Cint
+
+function SCS.scs_set_default_settings(
+    ::Type{SCS.GpuIndirectSolver},
+    stgs::SCS.ScsSettings{I},
+) where {I<:Cint}
+    return @ccall gpuindirect.scs_set_default_settings(
+        stgs::Ref{SCS.ScsSettings{I}},
+    )::Cvoid,
```
odow, line 23:
```suggestion
    )::Cvoid
```

odow:
@kalmarek I'm trying to sort it so that we support `@requires` on 1.6 and the package extensions on 1.9. 

We'll still have to do a breaking change to the use package loading, but that's okay. When 1.9 becomes the LTS, we can drop the requires part.

README.md
```diff
@@ -129,70 +129,57 @@ for other options.
...
+true
 ```
 
 The `MKLDirectSolver` is available on `Linux x86_64` platform only.
 
 ### SCS with Sparse GPU indirect solver (CUDA only)
 
-To enable the indirect linear solver on GPU one needs to load `CUDA_jll`
-**before** `SCS`:
+**SCS v2.0 introduced a breaking change. You now need to use `SCS_GPU_jll`
```
odow, line 170:
```suggestion
**SCS.jl v2.0 introduced a breaking change. You now need to use `SCS_GPU_jll`
```

README.md
```diff
@@ -129,70 +129,57 @@ for other options.
...
+using JuMP, SCS
+model = Model(SCS.Optimizer)
+set_attribute(model, "linear_solver", SCS.IndirectSolver)
+```
 
 ### SCS with MKL Pardiso linear solver
 
-To enable the MKL Pardiso (direct sparse) solver one needs to load `MKL_jll`
-**before** `SCS`:
+**SCS v2.0 introduced a breaking change. You now need to use `SCS_MKL_jll`
```
odow, line 149:
```suggestion
**SCS.jl v2.0 introduced a breaking change. You now need to use `SCS_MKL_jll`
```

kalmarek (changes requested):
Looks great!

src/SCS.jl
```diff
@@ -6,45 +6,41 @@
...
+            SCS_GPU_jll = "af6e375f-46ec-5fa0-b791-491b0dfa44a4",
+            include("../ext/SCSSCS_GPU_jllExt.jl"),
+        )
+        Requires.@require(
+            SCS_MKL_jll = "3f2553a9-4106-52be-b7dd-865123654657",
+            include("../ext/SCSSCS_MKL_jllExt.jl"),
+        )
+    end
+    return
+end
```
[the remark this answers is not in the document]
odow: Yeah that didn't work

odow:
cc @blegat you should take a look at this. Nominally it's a breaking change, but it affects only the lazy loading of the MKL and GPU solvers.

kalmarek (approved):

blegat (approved):

PR #293, opened 2024-07-05
add scs-3.2.6 jlls

kalmarek:


kalmarek:
@odow there's one (random numerical noise ?) fail on macos. If you think it's ok, I'm fine.

odow:
To clarify: the GPU fail is still expected?

kalmarek:
It is, it's been failing since CUDA-12.4 as per https://github.com/jump-dev/SCS.jl/issues/291#issuecomment-2213372992

PR #308, opened 2025-02-21
Add PrecompileTools.jl

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/2643


blegat:
What's the effect on this benchmark https://github.com/jump-dev/MathOptInterface.jl/pull/2643#issuecomment-2672561927 ?

odow:
14s to 4s

PR #310, opened 2025-05-09
Use MOI.Utilities.num_rows

blegat:
When I wrotet this hacky direct access to the `num_rows` field, there wasn't this `num_rows` function yet

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -384,8 +384,8 @@ function MOI.optimize!(
         P,
         Ab.constants.b,
         c,
-        MOI.Utilities.num_rows(Gh.sets, MOI.Zeros),
-        MOI.Utilities.num_rows(Gh.sets, MOI.Nonnegatives),
+        MOI.Utilities.num_rows(Ab.sets, MOI.Zeros),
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -384,8 +384,8 @@ function MOI.optimize!(
         P,
         Ab.constants.b,
         c,
-        Ab.sets.num_rows[1],
-        Ab.sets.num_rows[2] - Ab.sets.num_rows[1],
+        MOI.Utilities.num_rows(Gh.sets, MOI.Zeros),
```
odow, line 387:
```suggestion
        MOI.Utilities.num_rows(Ab.sets, MOI.Zeros),
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -384,8 +384,8 @@ function MOI.optimize!(
...
+        MOI.Utilities.num_rows(Gh.sets, MOI.Nonnegatives),
```
odow, line 388:
```suggestion
        MOI.Utilities.num_rows(Ab.sets, MOI.Nonnegatives),
```

PR #311, opened 2025-08-21
low level support for the complex psd cone

araujoms:
The 3.2.8 release of SCS added support for the complex PSD cone. Here I'm adding support for accessing it via the low-level interface. I'll add MOI support in a different PR, since @odow likes PRs to be as small as possible.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -10,6 +10,6 @@ steps:
           codecov: true
     agents:
       queue: "juliagpu"
-      cuda: "12"
+      cuda: "*"
     if: build.message !~ /\[skip tests\]/
     timeout_in_minutes: 60


odow (commented):
Your issue #312 implies that other cones were added. Do we not need to account for them also?

odow (approved):

test/test_problems.jl
```diff
@@ -440,7 +441,7 @@ function feasible_basic_conic(T)
     A = SparseMatrixCSC(m, n, colptr .+ 1, rowval .+ 1, vec(values))
     P = spzeros(n, n)
 
-    sol = scs_solve(T, m, n, A, P, b, c, z, l, bu, bl, q, s, ep, ed, Float64[])
+    sol = scs_solve(T, m, n, A, P, b, c, z, l, bu, bl, q, s, cs, ep, ed, Float64[])
```
odow, line 444: This is a breaking change to the SCS API???

odow: I merged because I don't know if many people are using the low-level API, and we can keep the MOI wrapper non-breaking. But I wonder if we should think about how to make this non-breaking

odow: I'll make a separate PR

araujoms: It's easy to make it non-breaking, just add a `scs_solve` with the old arguments calling the new one.

But `scs_solve` is not exported, and when @kalmarek added previous cones here he didn't make it non-breaking, so I decided to follow his example.

PR #315, opened 2025-08-23
add MOI interface for complex PSD cone

araujoms:
I've added a MOI interface for the complex PSD cone according to the design @blegat suggested [here](https://github.com/jump-dev/MathOptInterface.jl/issues/2597#issuecomment-2564271003), adding the sets `ScaledComplexPSDCone` and `ComplexPositiveSemidefiniteConeTriangle` and using the chain of bridges
`MOI.HermitianPositiveSemidefiniteConeTriangle` -> `ComplexPositiveSemidefiniteConeTriangle` -> `MOI.Scaled{ComplexPositiveSemidefiniteConeTriangle}` -> `ScaledComplexPSDCone`.

Still missing is the middle part `ComplexPositiveSemidefiniteConeTriangle` -> `MOI.Scaled{ComplexPositiveSemidefiniteConeTriangle}`, which I don't know how to do, but I expect it will be rather simple. I added the functions `set_dot` and `set_coefficients` for `ComplexPositiveSemidefiniteConeTriangle`, which I assume are necessary for `MOI.Scaled` to work.

### DIFF
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -61,7 +61,6 @@ function _test_runtests(linear_solver)
             # Unexpected failures:
             #   TODO(odow): looks like a tolerance issue?
             "test_linear_add_constraints",
-            "test_conic_HermitianPositiveSemidefiniteConeTriangle_2",
             # Expected test failures:
             #   TODO(odow): get not supported for primal/dual starts
             "test_model_ModelFilter_AbstractConstraintAttribute",


araujoms:
That was it, thanks a lot.

blegat (approved):
Looks good to me, the buildkit failure is unrelated

kalmarek:
@araujoms the only question that comes to my mind is: Is the interface/bridge really covered by the tests? I can only see a set of tests __removed__ in this PR :D

PR #330, opened 2025-10-20
Remove dependency on MutableArithmetics

odow:


araujoms:
Before your change
```julia
julia> using JuMP, SCS

julia> model = Model();

julia> set = SCS.ComplexPositiveSemidefiniteConeTriangle(50);

julia> @variable(model, x[1:2500]);

julia> @variable(model, y[1:2500]);

julia> @time MOI.Utilities.set_dot(x,y,set);
  0.370305 seconds (2.31 M allocations: 113.229 MiB, 4.22% gc time, 99.87% compilation time)

julia> @time MOI.Utilities.set_dot(x,y,set);
  0.000867 seconds (19.64 k allocations: 1.191 MiB)
```
After your change:
```julia
julia> using JuMP, SCS
Precompiling SCS finished.
  1 dependency successfully precompiled in 22 seconds. 56 already precompiled.

julia> model = Model();

julia> set = SCS.ComplexPositiveSemidefiniteConeTriangle(50);

julia> @variable(model, x[1:2500]);

julia> @variable(model, y[1:2500]);

julia> @time MOI.Utilities.set_dot(x,y,set);
  0.295810 seconds (291.24 k allocations: 431.760 MiB, 45.06% gc time, 39.39% compilation time)

julia> @time MOI.Utilities.set_dot(x,y,set);
  0.111543 seconds (106.55 k allocations: 422.962 MiB, 43.15% gc time)
```

odow:
What about an example where it isn't called directly? You're passing in JuMP variables. In MOI, it's only really used to compute the dual objective value, and it will always have `Float64` as the input type.

I'd be willing to just lock in many of the `T` types in the bridges for `Float64`. We're not trying to be generic. These sets and bridges have a very specific purpose.

araujoms:
No, I'm not going to waste my time writing more examples. Perhaps you shouldn't change code around for no benefit? I wrote the function the way it's done in MOI. And it will certainly be used for numerical types other than `Float64`, there are several solvers which can handle `BigFloat`.

odow:
These functions aren't even covered by the tests:

https://app.codecov.io/gh/jump-dev/SCS.jl/blob/master/src%2FMOI_wrapper%2Fsets%2FComplexPositiveSemidefiniteConeTriangle.jl

araujoms:
The point of writing this set was to incorporate it into MOI as the replacement for the complex PSD cone, as suggested by @blegat. It's supposed to be generic.

odow:
We already have too many PSD cones in MOI. I don't want to add another. It can stay in SCS.jl.

araujoms:
The point is not to add another set, but to replace one.

PR #333, opened 2025-12-14
Implement default_cache

blegat:
Otherwise, this isn't used by JuMP so we lose most of the performance advantage of using `MatrixOfConstraints`

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -495,7 +495,7 @@ end
 function MOI.optimize!(dest::Optimizer, src::MOI.ModelLike)
     linear_solver = get(dest.options, :linear_solver, DirectSolver)
     T = scsint_t(linear_solver)
-    cache = MOI.Utilities.UniversalFallback(OptimizerCache())
+    cache = MOI.Utilities.UniversalFallback(OptimizerCache{T}())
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -155,6 +155,12 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     Optimizer() = new(nothing, MOISolution(), false, Dict{Symbol,Any}())
 end
 
+function MOI.default_cache(::Optimizer, ::Type{Cdouble})
```
odow:
```suggestion
function MOI.default_cache(dest::Optimizer, ::Type{Cdouble})
```

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -488,7 +495,7 @@ end
 function MOI.optimize!(dest::Optimizer, src::MOI.ModelLike)
     linear_solver = get(dest.options, :linear_solver, DirectSolver)
     T = scsint_t(linear_solver)
-    cache = MOI.Utilities.UniversalFallback(OptimizerCache{T}())
+    cache = MOI.Utilities.UniversalFallback(OptimizerCache())
```
odow:
```suggestion
    cache = MOI.Utilities.UniversalFallback(OptimizerCache{T}())
```

PR #334, opened 2025-12-28
update SCS_jll to 300.200.1000

araujoms:
has a dramatic speedup for real SDPs due to switching `syev` to `syevr`.


araujoms:
My pleasure. Note that I don't have commit rights to this repository.

PR #360, opened 2026-09-07
Fix order of the sl_n and sl_k pointers in ScsCone constructor

odow:
Looks like it's just a typo from https://github.com/jump-dev/SCS.jl/pull/327, but I still need to make a proper test.

Upstream uses 

https://github.com/cvxgrp/scs/blob/35b491dfdead5e4bab595f48f6c9cb2aab868dc5/include/scs.h#L174-L176

image


odow:
@araujoms okay with this? Or were you passing the arguments reversed in `scs_solve`?

araujoms:
Me? I never used this cone. Change seems correct though.

PR #363, opened 2026-09-21
bump scs to 3.3.1

araujoms:


araujoms:
It's randomly segfaulting, no idea why.

bodono:
The segfault is a struct layout mismatch in `ScsInfo`. The PR updated `ScsSettings` for 3.3, but `ScsInfo` also changed.

SCS 3.3.0 added an `AaStats aa_stats` field to the C `ScsInfo` struct, between `accepted_accel_steps` and `lin_sys_time` (cvxgrp/scs@d7098d0f, "Expose Anderson acceleration stats in solve info"). It is defined in [`include/aa_stats.h`](https://github.com/cvxgrp/scs/blob/3.3.1/include/aa_stats.h) and used at [`include/scs.h#L236`](https://github.com/cvxgrp/scs/blob/3.3.1/include/scs.h#L236). The Julia `ScsInfo{T}` in `c_wrapper.jl` still has the 3.2 layout.

`AaStats` is eight `scs_int` plus two `scs_float`, so with `Clonglong` that is 80 bytes. Julia allocates a 400-byte object for `ScsInfo{Int64}` and passes its pointer to `scs_solve`, but C thinks the struct is 480 bytes. At the end of every solve C writes `aa_stats`, `lin_sys_time`, `cone_time` and `accel_time` past the end of the Julia object, clobbering whatever the GC placed next to it. The crash surfaces later, which is why it looks random. The CI log matches: the Ubuntu 1.10 job dies in `jl_gc_pool_alloc_inner` on the next array allocation right after a solve prints its summary.

Fix: insert the missing fields into the Julia struct, after `accepted_accel_steps` and before `lin_sys_time`:

```julia
    rejected_accel_steps::T
    accepted_accel_steps::T
    # AaStats aa_stats (new in SCS 3.3)
    aa_iter::T
    aa_n_accept::T
    aa_n_reject_lapack::T
    aa_n_reject_rank0::T
    aa_n_reject_nonfinite::T
    aa_n_reject_weight_cap::T
    aa_n_safeguard_reject::T
    aa_last_rank::T
    aa_last_aa_norm::Cdouble
    aa_last_regularization::Cdouble
    lin_sys_time::Cdouble
    cone_time::Cdouble
    accel_time::Cdouble
```

Flattening the nested struct is fine since it only holds scalars, so the layout is identical. Alternatively define an `AaStats{T}` immutable struct and embed it as a single field.

I diffed the other structs against the 3.3.1 headers: `ScsSettings` in the PR matches (including `adaptive_diag_scale` and the three new acceleration fields), and `ScsData`, `ScsCone` and `ScsSolution` are unchanged from 3.2. The `SPECTRAL_TIMING_FLAG` fields at the tail of `ScsInfo` are off by default so they don't need mirroring.

araujoms:
Thanks, that was it.

odow:
Is this ready to merge @araujoms?

araujoms:
Yes.

